PR #3, opened 2016-10-14
Fix Travis Mac build

mkuron:


hmenke:
According to the log it fails because `boost` and `cmake` are already installed.  The last line however reads `/Users/travis/build.sh: line 109: shell_session_update: command not found`

mkuron:
The last line doesn't matter, that's a bug in the Travis build environment. I have removed the unnecessary packages and we'll see if that fixes it.

PR #63, opened 2017-06-06
[python] Version dialog in GUI

### LINKED ISSUE #62: Show version number in the GUI
It would be nice if the user could get the version of the installed pairinteraction software in the GUI.

hmenke:
Fixes #62 

The dialog can be opened from the menu bar <kbd>Help > About</kbd>

version


hmenke:
@seweber Is this what you had in mind?

seweber:
@hmenke Yes, exactly. Thanks!

PR #95, opened 2019-01-24
FEAST eigensolver, Python wheels

seweber:


libpairinteraction/SystemBase.h
```diff
@@ -516,35 +521,128 @@ class SystemBase {
...
+        // Note that fpm[0]=1 enables terminal output and fpm[26]=1 enables a matrix checker.
+
+        MKL_INT fpm[128];
+        feastinit(fpm);
+        fpm[0] = 1;
+        fpm[1] = 6;
+        fpm[26] = 0;
+        fpm[3] = 5;
+        if (threshold != std::numeric_limits<double>::max()) {
+            fpm[2] = std::min(std::round(-std::log10(threshold)),
```
hmenke: This will blow up if a user sets `threshold = 0`.  Maybe use
```cpp
std::log10(1 + threshold) // potential problem for -1
```
or
```cpp
std::log10(threshold == 0 ? 1 : threshold) // extra branch
```

libpairinteraction/SystemBase.h
```diff
@@ -516,35 +521,128 @@ class SystemBase {
...
+        }
+
+        // Do the diagonalization
+        char uplo = 'F';                    // full matrix is stored
+        MKL_INT n = hamiltonian.rows();     // size of the matrix
+        MKL_INT info;                       // will contain return codes
+        double epsout;                      // will contain relative error
+        MKL_INT loop;                       // will contain number of used refinement
+        MKL_INT m;                          // will contain the number of eigenvalues
+        double *res = new double[m0];       // will contain the residual errors
```
hmenke: Don't use `new` and `delete`.  Just use a `std::vector`.

testsuite/feast.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class FeastTest(unittest.TestCase):
+
+    def setUp(self):
+        # Set up cache
+        self.cache = pi.MatrixElementCache()
+
+    def test_diagonalization_full(self):
+        if not pi.mkl_enabled:
```
hmenke: You can use the [`@unittest.skipIf`](https://docs.python.org/3.6/library/unittest.html#unittest.skipIf) decorator for the `FeastTest` class.  This will also mark the test properly as skipped.

PR #100, opened 2019-09-15
Julia interface

atombear:
this pr has basic functionality - not all methods are represented, and I will add them in the next week. This shows the general structure of the code, and any early suggestions would be much appreciated. I need to combine the two CMake files, presumably anticipating a -WJULIA flag.


atombear:
@hmenke thank you for your kind words!

I have merged all the cmake functionality, and I think everything is working. I will extend some methods in the next few days, but otherwise, this is ready for review and merge, unless there are other tests in specific you would like me to add.

hmenke:
I started adding Julia to the Docker containers.  It might take a while before everything is up and running there.  I also need instructions how to preinstall JlCxx such that it can be found by the C++ compiler.

pairinteraction/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ if(MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
 endif()
 
+add_compile_options(-Wno-unused-local-typedefs)
```
hmenke: This currently breaks the MSVC build.  Please remove it and append it to this:
```cmake
set_source_files_properties(${pairinteraction_SRCS} PROPERTIES COMPILE_FLAGS "-Wall -Wextra -pedantic -Werror")
```

atombear: done

atombear:
@hmenke for linux install - CxxWrap must be installed through the standard Julia package manager. The installation will come with all the relevant cpp files.

for testing, however - there is a problem I do not understand that I am trying to get help with.

hmenke:
I've updated the Docker images (#101) to install a Julia interpreter and the CxxWrap package.  Once that PR is merged you can rebase onto `master` and integrate the Julia stuff into the CI.

atombear:
excellent - it should also need SparseArrays, Test, and SQLite

atombear:
@hmenke hello, i am inquiring into the status of this pr. is there anything i can do to help? i'm sorry if i missed any signal.

hmenke:
I'm waiting for @seweber to approve #101 so that we have Julia available in the Docker images, so we can actually test the interface.  Also the Windows build is currently failing, which has to be fixed and @seweber is the expert for that.

seweber:
Thanks for the very neat work also from my side!

@hmenke I though that #101 was still afoot. From my view, we could directly add the Julia packages `SparseArrays`, `Test`, and `SQLite` to the Docker images as well. Concerning the problem with the Windows build: if I remember correctly, this is a weird bug in the interplay between AppVeyor and Miniconda. The current version of AppVeyor somehow tends to forget that we would like to use the Python 3 version of Miniconda. I worked around this bug in [5be7d89](https://github.com/pairinteraction/pairinteraction/pull/99/commits/5be7d893bde360df767c7c9b9d6d334528664e47). Cherry-picking this commit should fix the issue. Alternatively, we could merge #99 before merging #101. I would prefer this. There were some other small issues which were fixed in this pull request as well.

hmenke:
@seweber I've added the required Julia packages to the Docker images in https://github.com/pairinteraction/pairinteraction-build-dependencies/commit/12bf0a0d5a6a9044acea5da035dad171d509c08f, but as you know Docker Hub the build will probably take until the end of the day.

atombear:
I think I understand the first issue: I need to edit the make file.
I would like to be pointed in the right direction for the second issue.
I do not know how to start with the third issue :)

pairinteraction/CMakeLists.txt
```diff
@@ -48,20 +48,39 @@ configure_file(version.h.in ${CMAKE_CURRENT_BINARY_DIR}/version.h)
...
 file(GLOB pairinteraction_SRCS *.cpp)
 list(REMOVE_ITEM pairinteraction_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
+list(REMOVE_ITEM pairinteraction_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/julia_hooks_pi.cpp)
 
 add_library(pireal SHARED ${pairinteraction_SRCS} ${wignerSymbols_SRCS})
 add_library(picomplex SHARED ${pairinteraction_SRCS} ${wignerSymbols_SRCS})
 set_target_properties(picomplex PROPERTIES COMPILE_DEFINITIONS "USE_COMPLEX")
 
+if(WITH_JULIA)
+  execute_process(COMMAND locate JlCxxConfig.cmake OUTPUT_VARIABLE JlCxxConfig_LOC)
```
hmenke: `locate JlCxxConfig.cmake` won't work on Windows and on Linux it requires `mlocate` to be installed and the database to be up-to-date.  Instead something like
```suggestion
  find_file(JlCxxConfig_LOC NAMES JlCxxConfig.cmake)
```
should be used, but there you have to provide one or more paths where to search for `JlCxxConfig.cmake`, because it is not located in any of the standard paths.

hmenke: It's possible to get the path using
```console
$ julia -e 'using CxxWrap; print(joinpath(dirname(CxxWrap.jlcxx_path), "cmake", "JlCxx"))'
```

hmenke: We could set
```bash
export CMAKE_PREFIX_PATH="$(julia -e 'using CxxWrap; print(joinpath(dirname(CxxWrap.jlcxx_path), "cmake", "JlCxx"))')${CMAKE_PREFIX_PATH:+:${CMAKE_PREFIX_PATH}}"
```
although this doesn't really belong inside the Docker container.

seweber: There might be a simpler solution, see e.g. https://github.com/JuliaInterop/CxxWrap.jl/issues/115#issuecomment-437347226. We could try something like `get_target_property(JlCxx_location JlCxx::cxxwrap_julia LOCATION)`. However, I do not completely understand why we need this path, i.e. why we have to set `CMAKE_INSTALL_RPATH`?

hmenke: You misunderstood.  Finding `JlCxxConfig.cmake` is required for `find_package(JlCxx)` to work.

But you are right, that the code following `find_package(JlCxx)` can be simplified with the stuff you linked.

seweber: I see. Could we get the Julia executable by calling `find_program` in CMake and use `execute_process` to call the executable and run your Julia code to obtain the path?

hmenke:
Looks like Homebrew has removed *all* Sierra bottles now.  Maybe we should update to `xcode: 9.4` in Travis.

seweber:
Yes, I agree. I made a pull request where I removed the llvm dependency in addition, see #102.

seweber:
I have now enabled Julia for the Linux builds, using `-DWITH_JULIA=On`. Finding Julia via CMake seems to work nicely :-). However, there are some issues in the later stages of the built process, see the [travis log](https://travis-ci.org/pairinteraction/pairinteraction/builds/608199441?utm_source=github_status&utm_medium=notification).

hmenke:
Here are the main errors:

`image=static-analysis` https://travis-ci.org/pairinteraction/pairinteraction/jobs/608199442#L436-L1638

> Fails with clang-tidy.  Error message too long to show here.

---

`image=ubuntu` https://travis-ci.org/pairinteraction/pairinteraction/jobs/608199446#L522-L528
```
  dpkg-shlibdeps: error: cannot find library libcxxwrap_julia.so.0 needed by
  ./usr/lib/libpicomplex.so (ELF format: 'elf64-x86-64' abi:
  '0201003e00000000'; RPATH: '')

  dpkg-shlibdeps: error: cannot find library libcxxwrap_julia.so.0 needed by
  ./usr/lib/libpireal.so (ELF format: 'elf64-x86-64' abi: '0201003e00000000';
  RPATH: '')
```

---

`image=opensuse` https://travis-ci.org/pairinteraction/pairinteraction/jobs/608199447#L431
```
/travis/build/pairinteraction/unit_test/sqlite: error while loading shared libraries: libjulia.so.1: cannot open shared object file: No such file or directory
```

seweber:
I have thought a bit about the last two errors:
* The error in `image=ubuntu` might be caused by a missing/wrong rpath. Maybe, the problem is that `set(CMAKE_INSTALL_RPATH [...])` is called after the `add_library` command. If I remember correctly, I read somewhere that it should be called before that command (as it is done in [this CMake example](https://github.com/JuliaInterop/libcxxwrap-julia/blob/master/testlib-builder/src/testlib/CMakeLists.txt)). However, I wonder whether using `${JlCxx_DIR}/../` as rpath breaks the portability of the created package. It might be better to statically link to `libcxxwrap_julia` or include the dynamic library in the package if possible. If nothing else works, we could think about adding the path to `libcxxwrap_julia` to `LD_LIBRARY_PATH`. However, then all user of the library would also have to modify their `LD_LIBRARY_PATH`. Thus, I am not in favour of this solution. How is it done in other projects using cxxwrap?
* I do not understand the error in `image=opensuse`. According to the log, CMake recognises `libjulia.so` in the directory `/usr/lib/`. But the pairinteraction library tries to load `libjulia.so.1`. Is this behaviour expected? We could switch on verbose output in CMake and study the linker command.

hmenke:
I think what is being done right now is just wrong.  We should not link the core `libpireal.so` and `libpicomplex.so` with `libjulia.so`.  After all we are not linking them with `libpython.so` either.  The proper design would be the same as for the Python interface, i.e. have a separate library which includes the Julia hooks and links to `libjulia.so`, which is then only used within the Julia interpreter.

seweber:
I completely agree that this would be the cleaner approach and we should change it. But I don't expect that this fixes the issues.

PairInteraction.jl/src/PairInteraction.jl
```diff
@@ -0,0 +1,34 @@
+module PairInteraction
+
+import Base.^
+import Base.string
+using CxxWrap
+
+pairinteraction_dir = dirname(@__FILE__)
+build_dir = joinpath(splitdir(splitdir(pairinteraction_dir)[1])[1], "build", "pairinteraction")
```
hmenke: I replaced `splitpath` since that is only available in Julia 1.1 but the Docker containers ship 1.0.5 (which I would prefer to keep because it's the LTS variant).

PairInteraction.jl/test/test_quantum_defect.jl
```diff
@@ -0,0 +1,21 @@
+# TestQuantumDefect
+using Test
+using SQLite
+using PairInteraction
+
+# test_comparison
+qd = PairInteraction.QuantumDefect("Rb", 78, 1, 0.5)
+
+sqldb = SQLite.DB("pairinteraction/databases/quantum_defects.db")
```
hmenke: The working directory for the tests is now the testsuite directory, so we can use a relative path for the database.

pairinteraction/CMakeLists.txt
```diff
@@ -48,20 +48,42 @@ configure_file(version.h.in ${CMAKE_CURRENT_BINARY_DIR}/version.h)
...
 
+if(WITH_JULIA)
+  find_package(JuliaInterp REQUIRED)
+  execute_process(
+    COMMAND "${Julia_EXECUTABLE}" -e "using CxxWrap; print(joinpath(dirname(CxxWrap.jlcxx_path), \"cmake\", \"JlCxx\"))"
+    OUTPUT_VARIABLE JlCxx_DIR
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  find_package(JlCxx REQUIRED)
+  #set(CMAKE_INSTALL_RPATH "${JlCxx_DIR}/../")
+  #set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
hmenke: Can we remove this now?

pairinteraction/CMakeLists.txt
```diff
@@ -48,20 +48,42 @@ configure_file(version.h.in ${CMAKE_CURRENT_BINARY_DIR}/version.h)
...
+
 target_link_libraries(pairinteraction-real    pireal)
 target_link_libraries(pairinteraction-complex picomplex)
+
 if(MSVC)
   add_definitions(/W3 /DNOMINMAX /DWIN32_LEAN_AND_MEAN)
   #set_source_files_properties(${pairinteraction_SRCS} PROPERTIES COMPILE_FLAGS "/W3 /DNOMINMAX /DWIN32_LEAN_AND_MEAN")
 else()
-  set_source_files_properties(${pairinteraction_SRCS} PROPERTIES COMPILE_FLAGS "-Wall -Wextra -pedantic -Werror")
+  set_source_files_properties(${pairinteraction_SRCS} PROPERTIES COMPILE_FLAGS "-Wall -Wextra -pedantic -Werror -Wno-unused-local-typedefs")
```
hmenke: `-Wno-unused-local-typedefs` can be removed as well I guess.  If at all, it should be
```
set_source_files_properties(julia_hooks_pi.cpp PROPERTIES COMPILE_FLAGS "-Wall -Wextra -pedantic -Werror -Wno-unused-local-typedefs")
```

seweber:
Excellent! Should we provide binary builds of the Julia bindings?

pairinteraction/CMakeLists.txt
seweber: I would remove it.

hmenke:
On my personal GitHub there is also a further commit where I have encapsulated the Julia bindings a bit more nicely into their own directory (and renamed it from `PairInteraction.jl` to `parinteraction.jl`).

Binaries should be provided as a proper Julia package and ideally published on https://pkg.julialang.org/ so that people can do simple `Pkg.add` to install pairinteraction.

Docs on how to build packages are located here: https://julialang.github.io/Pkg.jl/v1/creating-packages/

seweber:
I my opinion it would be perfect to add your commit https://github.com/hmenke/pairinteraction/commit/7d68ea5b3573439432f12dd8a151bb6279666620 to the pull request.

seweber:
The easiest way to provide a Julia package seems to create an extra repository `pairinteraction.jl` for the package. If the repository contains the proper files, the package can be installed by simply executing `pkg> add https://github.com/pairinteraction/pairinteraction.jl`. Then, we can also make use of the [Registrator](https://github.com/JuliaRegistries/Registrator.jl) GitHub app to put the package to the general Julia registry, so that it can be installed via `pkg> add pairinteraction`. @hmenke Do you agree that we should create a new repository `pairinteraction.jl` where we put the Julia bindings? The repository could be automatically populated by Travis CI as we already do it with the gh-pages branch. 

One thing which perhaps requires some further thinking is how to provide all the necessary binary files. The binary files are different for different operating systems. Unfortunately, it seems to be not possible to provide different Julia packages for different operating systems. Instead, we would have to write one single Julia package that checks the operating system and then looks for the correct binary files. There are different possibilities of how this could be achieved in practice. For example, we could provide a collection of the necessary binary files in our GitHub releases section for different operating systems. Then, the Julia package could download the proper binary files from there. Alternatively, we could try to include the library for the Julia binding to the installation files of the pairinteraction software. Then, the Julia package could check whether the pairinteraction software has been installed. If yes, it can take all necessary binaries from this installation. If no, it can prompt a message that pairinteraction should be installed. Another approach would be to make use of the [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl), but this wouldn't work for Windows and wouldn't allow putting our package to the general Julia registry if I understand things correctly. @atombear Maybe, you can also take a look.

seweber:
I have searched the [general Julia registry](https://github.com/JuliaRegistries/General) for packages using `CxxWrap`, so that we can get some inspiration of how to make the Julia package.
* Packages using `BinaryProvider` in addition, i.e. packages which download binary dependencies created via `BinaryBuilder`: 
  * https://github.com/JuliaOpt/SDPA.jl
  * https://github.com/JuliaReinforcementLearning/ViZDoom.jl
  * https://github.com/jstrube/FastJet.jl
  * https://github.com/oscar-system/Polymake.jl
  * https://github.com/jstrube/LCIO.jl

* Packages not using `BinaryProvider`:
  * https://github.com/JuliaFEM/MFrontInterface.jl
  * https://github.com/xtensor-stack/Xtensor.jl

PR #107, opened 2020-01-28
Remove CodeCov

hmenke:
I don't think we are using CodeCov to its full potential and right now it's only causing apparent build failures, when in reality the build passes.  Since we have our own coverage system, I think CodeCov is mostly redundant and can be removed.


hmenke:
Also removed CodeCov from repo settings.

PR #103, opened 2019-12-22
Remove the Boost dependency

hmenke:
- [x] `boost::hash_combine` and others.
- [x] Boost.Format was replaced by the fmt library.
- [x] Boost.Lexical_Cast was removed. I just wrote all necessary overloads by hand.
- [x] Boost.Test was replaced by Catch2.
- [x] Boost.Serialization was replaced by cereal.  Unfortunately, cereal does not support a nice flat text format like Boost.Serialization.
- [x] Boost.Program_options was removed.  Our argument parser is embarrassingly simple anyway.
- [x] Boost.PropertyTree was replaced by rapidjson which comes as part of cereal.

These headers still remain:
```console
$ git grep '<boost/' | sed -e 's/^.*<\(.*\)>/\1/g' | sort | uniq
boost/algorithm/hex.hpp
boost/core/ignore_unused.hpp
boost/filesystem.hpp
boost/iostreams/device/array.hpp
boost/iostreams/stream.hpp
boost/iterator/iterator_facade.hpp
boost/math/special_functions/binomial.hpp
boost/math/special_functions/factorials.hpp
boost/multi_index_container.hpp
boost/multi_index/hashed_index.hpp
boost/multi_index/member.hpp
boost/multi_index/random_access_index.hpp
boost/optional.hpp
boost/tokenizer.hpp
boost/uuid/uuid_generators.hpp
boost/uuid/uuid.hpp
boost/variant.hpp
```

PR #110, opened 2020-02-27
[cmake] Move submodules to libs folder

hmenke:
Also the Eigen GitHub mirror is going to be retired soon, so I though this is a good opportunity to point the submodule at the new mirror.

This change is also part of #103, so if we merge #103 first, this one is obsolete.


hmenke:
You might have to run
```
git submodule update --init
```
on your local checkout after pulling this.

PR #116, opened 2020-07-05
Explicit specification of two-atom basis

seweber:
We can pass a vector containing pairs of indices of one-atom basis vectors to `SystemTwo.setOneAtomBasisvectors()`. The specified pairs of one-atom basis vectors are then combined to two-atom basis vectors.


pairinteraction/SystemTwo.cpp
```diff
@@ -225,170 +236,182 @@ void SystemTwo::initializeBasis() {
...
 
-            // Continue if the pair statet energy is not valid
-            double energy = std::real(system1.getHamiltonian().coeff(col_1, col_1) +
-                                      system2.getHamiltonian().coeff(col_2, col_2));
-            if (!checkIsEnergyValid(energy)) {
-                continue;
+    // Loop over the pairs
+    for (const auto &a : pairs) {
+        size_t col_1 = a[0];
+        size_t col_2 = a[1];
```
hmenke, line 256: Since we are using C++17 now, you can also use structured bindings for this type of loop:
```suggestion
    for (const auto [col_1, col_2] : pairs) {
```

pairinteraction/SystemTwo.cpp
```diff
@@ -152,6 +152,17 @@ void SystemTwo::setConservedMomentaUnderRotation(const std::set<int> &momenta) {
     }
 }
 
+void SystemTwo::setOneAtomBasisvectors(const std::vector<std::array<size_t, 2>> &indices) {
+    // Check that all indices are unique
+    std::vector<std::array<size_t, 2>> tmp(indices);
+    std::sort(tmp.begin(), tmp.end());
+    if (std::adjacent_find(tmp.begin(), tmp.end()) != tmp.end()) {
+        throw std::runtime_error("The indices are not unique.");
+    }
```
hmenke, line 161: Might be nice to say which indices are duplicate:
```suggestion
    if (auto const it = std::adjacent_find(tmp.begin(), tmp.end()); it != tmp.end()) {
        std::string const which = "{" + std::to_string((*it)[0]) + "," + std::to_string((*it)[1]) + "}";
        throw std::runtime_error("The indices are not unique: " + which);
    }
```

PR #118, opened 2020-07-07
Remove some Boost dependencies

hmenke:
This removes some of the Boost dependencies which are unnecessary and can be easily replaced by standard library functionality.  This is part of #103.

Bold entries require linkage.
- [x]  1. Boost.Lexical_Cast was removed. I just wrote all necessary overloads by hand.
- [x]  2. Boost.Iterator was removed and the iterator functions written by hand.  The iterator is no longer conforming but I don't think anybody cares.
- [x]  3. Boost.Core was removed.  `boost::ignore` was replaced by `(void)` cast.
- [x]  4. Boost.Optional could be replaced by `<optional>` in C++17
- [x]  5. Boost.Variant could be replaced by `<variant>` in C++17
- [x]  6. **Boost.Program_options** was removed.  Our argument parser is embarrassingly simple anyway.

### DIFF
--- a/pairinteraction/CMakeLists.txt
+++ b/pairinteraction/CMakeLists.txt
@@ -104,7 +104,7 @@ find_package(Sqlite3 REQUIRED)
 include_directories(SYSTEM ${SQLITE3_INCLUDE_DIR})
 list(APPEND LIBRARIES ${SQLITE3_LIBRARY})
 
-find_package(Boost COMPONENTS filesystem system program_options serialization REQUIRED)
+find_package(Boost COMPONENTS filesystem system serialization REQUIRED)
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 list(APPEND LIBRARIES ${Boost_LIBRARIES})
 if(MSVC)
--- a/pairinteraction/Cache.h
+++ b/pairinteraction/Cache.h
@@ -21,9 +21,8 @@
 #define CACHE_H
 
 #include <mutex>
+#include <optional>
 #include <unordered_map>
-
-#include <boost/optional.hpp>
 
 /** \brief Generic cache object
  *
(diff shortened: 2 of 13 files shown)


seweber:
Looks very good. As an additional check, I have used pairinteraction with the suggested changes and commit 6009551b24e4124f3f623f94baab7fdea8dfd9d2 on our institute's cluster. It runs stably and the run times are very similar as before. From my point of view, we can hence add commit 6009551b24e4124f3f623f94baab7fdea8dfd9d2 to the pull request and merge it.

hmenke:
Okay, 6009551 is queued. Just waiting for the tests to pass.

hmenke:
Windows CI fails but this has actually nothing to do with this PR. Searching the log for "error" reveals these three instances:

- `interface_complexPYTHON_wrap.obj : error LNK2019: unresolved external symbol __imp_PyInstanceMethod_New referenced in function SWIG_PyInstanceMethod_New [C:\projects\pairinteraction\build\pairinteraction\_picomplex.vcxproj]`

- `interface_realPYTHON_wrap.obj : error LNK2019: unresolved external symbol __imp_PyInstanceMethod_New referenced in function SWIG_PyInstanceMethod_New [C:\projects\pairinteraction\build\pairinteraction\_pireal.vcxproj]`

- `CUSTOMBUILD : 15215 error : Hidden import 'pkg_resources.py2_warn' not found [C:\projects\pairinteraction\build\pairinteraction_gui\pyinstaller.vcxproj]`

I think we can actually ignore the first two, because the `__imp_PyInstanceMethod_New` is a symbol defined by the Python interpreter, so it will be there once the module is loaded in Python.  To get rid of this we probably only need to explicitly link with `libpython.dll`.  This error actually does not seem to be fatal (probably for the aforementioned reason) and has already been there before.

The last error is the one that crashes the build.  This is unfortunately problematic, because it stems from 7ee77d83d0a885e9fe2db234e9b451604377da71 which I have introduced to work around a bug in recent `setuptools`.  Maybe this is now fixed in an updated version of `setuptools` but that also means that we have to update our toolchain.

PR #121, opened 2022-02-03
Add multipole Rydberg-atom-ion interaction

tripiti:
The changes provide the ability to calculate the multipolar interaction (monopole-dipole, monopole-quadrupole,..) between an ion and a Rydberg atom. The maximum order of the interaction is set by setRydIonOrder(maxorder) where maxorder=1 includes only the monopole-dipole, maxorder=2  includes monopole-dipole and monopole-quadrupole.

For the 45P state the calculations provide very similar results compared to Fig. 2 (d)  in A. Duspayev, X. Han, M. A. Viray, L. Ma, J. Zhao, and G. Raithel, Long-Range Rydberg-Atom–Ion Molecules of Rb and Cs, Phys. Rev. Research 3, 023114 (2021).

### DIFF
--- a/pairinteraction/SystemOne.cpp
+++ b/pairinteraction/SystemOne.cpp
@@ -32,12 +32,13 @@
 
 SystemOne::SystemOne(std::string species, MatrixElementCache &cache)
     : SystemBase(cache), efield({{0, 0, 0}}), bfield({{0, 0, 0}}), diamagnetism(true),
-      species(std::move(species)), sym_reflection(NA), sym_rotation({static_cast<float>(ARB)}) {}
+    charge(0), ordermax(0), distance(std::numeric_limits<double>::max()),
+    species(std::move(species)), sym_reflection(NA), sym_rotation({static_cast<float>(ARB)}) {}
 
 SystemOne::SystemOne(std::string species, MatrixElementCache &cache, bool memory_saving)
     : SystemBase(cache, memory_saving), efield({{0, 0, 0}}), bfield({{0, 0, 0}}),
-      diamagnetism(true), species(std::move(species)), sym_reflection(NA),
-      sym_rotation({static_cast<float>(ARB)}) {}
+      diamagnetism(true), charge(0), ordermax(0), distance(std::numeric_limits<double>::max()),
+      species(std::move(species)), sym_reflection(NA), sym_rotation({static_cast<float>(ARB)}) {}
 
 const std::string &SystemOne::getSpecies() const { return species; }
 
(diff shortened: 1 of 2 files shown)


seweber:
Thanks again for your pull request! As discussed, I took the pull request and added the example notebook that you have sent to me. I slightly reduced the accuracy of the calculation within the notebook, hoping that this makes the notebook to run fast enough to be also used in our test system. In addition, I merged our latest changes with the pull request. I pushed my changes to the branch  [pairinteraction:IonAtomInteraction](https://github.com/pairinteraction/pairinteraction/tree/IonAtomInteraction). You can push this branch to the branch of your pull request to add my commits. Alternatively, we can directly merge pairinteraction:IonAtomInteraction to the master branch of pairinteraction.

seweber:
Hmm, running the notebook takes still too much time for the continuous integration, over 900 seconds on GitHub Actions (although it runs on my local computer within 10 seconds). Maybe, we should not use this notebook for integration testing but write some small test cases.

hmenke:
The timeout occurs (unsurprisingly) only in the `debian` job, because this one is built without optimizations to obtain code coverage data for https://www.pairinteraction.org/pairinteraction/coverage/html/index.html

I will refactor the CMake to make it possible to also register notebooks that are only to be run in release mode. Please also keep in mind that the notebook tests are not unit tests. They only test whether the notebook can be executed but not the correctness of their results.

hmenke:
@seweber I don't know whether you've seen, but I enabled documentation building and notebook testing also for the `ubuntu` and `opensuse` jobs which build in release mode on the `master` branch. There is now an extra section in the `CMakeLists.txt` with notebooks that will only run in release mode to not hit the timeout:

https://github.com/pairinteraction/pairinteraction/blob/6bd256128a67e8f8f87d9a1744e3644dbc737dab/doc/sphinx/CMakeLists.txt#L138-L143

seweber:
Perfect, I now also added a small unit test that checks the energy shift due to atom-ion-interaction. From my point of view, we can merge the [pairinteraction:IonAtomInteraction](https://github.com/pairinteraction/pairinteraction/tree/IonAtomInteraction) branch to master and close the pull request.

hmenke:
Could you please rebase onto `master` instead of merging? While doing that please also drop https://github.com/pairinteraction/pairinteraction/commit/a6565c8ec4a936a60a2ae270b4fbfa9e31079518. There should be no need to increase the timeout.

seweber:
Yes, you are right. Now, the branch is cleaned up a bit.

hmenke:
LGTM :ok_hand: You can push that branch to master.

PR #125, opened 2022-08-05
macOS: 10.15 -> 11

hmenke:
In macOS 11 Big Sur `/usr/lib/libsqlite3.dylib` does not exist anymore. It's now a system library that will be resolved by the dynamic loader upon calling `dlopen("/usr/lib/libsqlite3.dylib")`. Essentially that path has become a special name without any actual file backing it.

Therefore checking for the existence of libraries should only be done for non-system dependencies.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -21,7 +21,7 @@ env:
 
 jobs:
   macos:
-    runs-on: macos-10.15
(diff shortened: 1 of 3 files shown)


seweber (approved):
Nice idea to use tmate to debug the build! The fix looks good.

PR #132, opened 2023-02-07
Sanitizers and Flatpak fixes

hmenke:
This enables support for ASan, UBSan and libstdc++ debug mode and fixes some of the errors that were found using these.

Further some problems that arose during Flatpak preparation are addressed:
- [x] The `python.cache` test assets have been downloaded and placed in the testsuite directory so that the test can work offline.
- [x] The software version can now be set from the command line so that configuring also works when not building from a full clone.
- [x] Add a QSplitter to the GUI to make settings and plot panels resizable relative to each other.


CMakeLists.txt
```diff
@@ -33,17 +33,25 @@ set(VERSION_CACHE "v17")
...
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if(GIT_DESCRIBE_RESULT EQUAL 0)
+      set(VERSION_SOFTWARE "${GIT_DESCRIBE_OUTPUT}" CACHE STRING "Set software version" FORCE)
+    else()
+      message(WARNING "Git failed to determine the version of the pairinteraction software.")
+      set(VERSION_SOFTWARE "v0.0.0" CACHE STRING "Set software version" FORCE)
+    endif()
+  else()
+    message(WARNING "Because Git was not found, the version of the pairinteraction software cannot be determined.")
+    set(VERSION_SOFTWARE "v0.0.0" CACHE STRING "Set software version" FORCE)
```
hmenke, line 53: We could also think about
```cmake
set(VERSION_SOFTWARE "${VERSION_SOFTWARE_DEFAULT}" CACHE STRING "Set software version" FORCE)
```
and further above do
```cmake
set(VERSION_SOFTWARE_DEFAULT "v0.9.8-alpha")
```
which has to be bumped manually with every release. This would always guarantee a more sensible version number but at the expense of some manual work at release time. I think we should also write a release checklist at some point.

seweber (approved):

PR #136, opened 2023-02-27
Sign change in wavefunctions.ipynb

frederic-hummel:
Changed the sign of the differential equation in the Numerov method section for consistency with Numerov integration step.

### DIFF
--- a/doc/sphinx/examples_python/wavefunctions.ipynb
+++ b/doc/sphinx/examples_python/wavefunctions.ipynb
@@ -104,7 +104,7 @@
     "Numerov's method can be used to solve differential equations of the form\n",
     "\n",
     "$$\n",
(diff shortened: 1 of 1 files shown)


hmenke:
Are you sure that's correct? According to [Wikipedia](https://en.wikipedia.org/wiki/Numerov%27s_method) Numerov's method solves differential equations of the form

Differential equation

hmenke (approved):

frederic-hummel:
Hi Henri,jup, I’m sure, it corresponds to the following equations in that section. Note that e.g. in the Numerov step the sign is opposite to the Wikipedia convention. That got me bamboozled for a while :)Best,FredericSent from my iPhoneOn 27. Feb 2023, at 08:05, Henri Menke ***@***.***> wrote:﻿
Are you sure that's correct? According to Wikipedia Numerov's method solves differential equations of the form

PR #137, opened 2023-03-03
Use correct platform cache and config dirs

### LINKED ISSUE #135: Use proper cache and config locations
Currently we dump all of the state of the pairinteraction GUI into `~/.pairinteraction`. This goes against all desktop specifications on all operating systems and should be fixed. Unfortunately this is platform dependent, so we need to distinguish:

- Config Directory:

    Linux: `$XDG_CONFIG_HOME/pairinteraction` or `~/.config/pairinteraction`
    Windows: `%APPDATA%\pairinteraction`
    MacOS: `~/Library/Preferences/pairinteraction`

- Cache directory:

    Linux: `$XDG_CACHE_HOME/pairinteraction` or `~/.cache/pairinteraction`
    Windows: `%LOCALAPPDATA%\pairinteraction`
    MacOS: `~/Library/Caches/pairinteraction`

hmenke:
Fixes #135

This should now use the correct locations for cache and config directories depending on the operating system. However, it needs to be tested for each platform that these are indeed correct.

### DIFF
--- a/pairinteraction_gui/pairinteraction/app.py
+++ b/pairinteraction_gui/pairinteraction/app.py
@@ -483,21 +483,50 @@ class MainWindow(QtWidgets.QMainWindow):
         self.path_cpp_complex = os.path.join(self.path_base, self.path_workingdir, "pairinteraction-complex")
         self.path_quantumdefects = os.path.join(self.path_base, self.path_workingdir, "databases/quantum_defects.db")
 
-        if os.name == "nt":
-            self.path_out = os.path.join(self.userpath, "pairinteraction/")
+        if sys.platform == "win32":
+            self.path_config = os.path.expandvars(r"%APPDATA%\pairinteraction")
+            if (
+                r"%APPDATA%" in self.path_config
+                or self.path_config == r"\pairinteraction"
+                or not os.path.isabs(self.path_config)
+            ):
+                self.path_config = os.path.expanduser(r"~\AppData\Roaming\pairinteraction")
+            self.path_cache = os.path.expandvars(r"%LOCALAPPDATA%\pairinteraction")
+            if (
+                r"%LOCALAPPDATA%" in self.path_cache
+                or self.path_cache == r"\pairinteraction"
+                or not os.path.isabs(self.path_cache)
+            ):
+                self.path_cache = os.path.expanduser(r"~\AppData\Local\pairinteraction")
+        elif sys.platform == "darwin":
+            self.path_config = os.path.expanduser(r"~/Library/Preferences/pairinteraction")
+            self.path_cache = os.path.expanduser(r"~/Library/Caches/pairinteraction")
         else:
-            self.path_out = os.path.join(self.userpath, ".pairinteraction/")
-        self.path_cache = os.path.join(self.path_out, "cache/")
+            self.path_config = os.path.expandvars(r"$XDG_CONFIG_HOME/pairinteraction")
+            if (
+                r"$XDG_CONFIG_HOME" in self.path_config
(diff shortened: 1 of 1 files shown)


pairinteraction_gui/pairinteraction/app.py
```diff
@@ -483,21 +483,34 @@ def __init__(self, parent=None):
...
 
         self.path_system_last = os.path.join(self.path_lastsettings, "lastsettings.sconf")
         self.path_plot_last = os.path.join(self.path_lastsettings, "lastsettings.pconf")
         self.path_view_last = os.path.join(self.path_lastsettings, "lastview.json")
         self.path_cache_last = os.path.join(self.path_lastsettings, "lastcache.json")
 
-        self.path_config = os.path.join(self.path_out, "conf.json")
-        self.path_version = os.path.join(self.path_out, "version.json")
+        self.path_config = os.path.join(self.path_config, "conf.json")
+        self.path_version = os.path.join(self.path_config, "version.json")
```
seweber, line 513: These lines set `self.path_version` to "[...]/conf.json/version.json", but it should be "[...]/version.json". To fix this bug, you could rename `self.path_out` to `self.path_preferences` instead of renaming it to `self.path_config`.

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -483,21 +483,34 @@ def __init__(self, parent=None):
...
         self.path_quantumdefects = os.path.join(self.path_base, self.path_workingdir, "databases/quantum_defects.db")
 
-        if os.name == "nt":
-            self.path_out = os.path.join(self.userpath, "pairinteraction/")
+        if sys.platform == "win32":
+            self.path_config = os.path.expandvars(r"%APPDATA%\pairinteraction")
+            if r"%APPDATA%" in self.path_config:
+                self.path_config = os.path.expanduser(r"~\AppData\Local\pairinteraction")
+            self.path_cache = os.path.expandvars(r"%LOCALAPPDATA%\pairinteraction")
+            if r"%LOCALAPPDATA%" in self.path_config:
```
seweber, line 491: Shouldn't it be if `r"%LOCALAPPDATA%" in self.path_cache`?

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -483,21 +483,34 @@ def __init__(self, parent=None):
         self.path_cpp_complex = os.path.join(self.path_base, self.path_workingdir, "pairinteraction-complex")
         self.path_quantumdefects = os.path.join(self.path_base, self.path_workingdir, "databases/quantum_defects.db")
 
-        if os.name == "nt":
-            self.path_out = os.path.join(self.userpath, "pairinteraction/")
+        if sys.platform == "win32":
+            self.path_config = os.path.expandvars(r"%APPDATA%\pairinteraction")
+            if r"%APPDATA%" in self.path_config:
+                self.path_config = os.path.expanduser(r"~\AppData\Local\pairinteraction")
```
seweber, line 489: Shouldn't it be `self.path_config = os.path.expanduser(r"~\AppData\Roaming\pairinteraction")`?

seweber:
I tried the GUI with Windows 11 and macOS Big Sur. The paths seem to be correct, except for my comments above. However, I encountered a bug under macOS at the startup of pairinteraction (`loadSettingsSystem` fails in line 1097 of "app.py" with `KeyError: '[mass] ** 0.5 / [length] ** 0.5 / [time]'` that is raised in line 132 of "unitmanagement.py"), but probably the bug is unrelated to the pull request.

seweber (approved):

PR #139, opened 2023-03-05
Continuous benchmarking

hmenke:
https://hmenke.github.io/pairinteraction/dev/bench/index.html

I hope I have covered all cases and combinations of gh-pages existing and not existing.


benchmarks/integration_bench.cpp
```diff
@@ -0,0 +1,73 @@
+#include "MatrixElementCache.h"
+#include "State.h"
+#include "SystemOne.h"
+#include "SystemTwo.h"
+#include "dtypes.h"
+#include "filesystem.h"
+
+#include <benchmark/benchmark.h>
+
+static void benchmark_integration(benchmark::State &state) {
```
hmenke, line 10: This benchmark is actually really poor. It only manages to run a single digit number of iteration (usually 6 to 8). To reduce variance benchmarks should do tens of thousands, better yet millions of iterations.

seweber: Because the code that is benchmarked is not very short, I would have expected some self-averaging effect. Thus, I am surprised by the [very high variance](https://hmenke.github.io/pairinteraction/dev/bench/index.html). I guess that the variance comes rather from the benchmarking environment being not stable. As a workaround, we could use [relative benchmarking](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks): First run the benchmark with a previous commit, then with the actual commit. Afterward, use [compare.py --dump_to_json](https://github.com/google/benchmark/blob/main/docs/tools.md#comparepy) to see the difference.

seweber: Note that a drawback of my proposal would be that it would be more complicated to use it with `github-action-benchmark`: I don't think that the JSON output of `compare.py` will be compatible with the format expected by this action. However, we could either write a script that rewrites the output to that format or discard this action and use some javascript library for making a nice histogram.

hmenke: I don't understand this reasoning. If there are only six samples then any outlier will strongly affect the mean and variance. This is a problem especially because the first iteration starts with a cold cache.

Relative benchmarking is performed by the github-action-benchmark which is what is used to trigger warnings for performance regressions.

seweber: In general, I agree. However, for the conducted benchmark, I wonder whether the huge variations can be really explained by the sample size. Have you tried it on your local computer with the same sample size? If yes, what have been typical variations there? My guess is that for each execution of GitHub actions, the virtual environment in which the script runs behaves a bit differently. The overall workload of the server where the virtual environment runs might vary, maybe even the hardware or software of the host system is not always the same. I just found a comment on the stability of the virtual environments in the documentation of [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark/tree/78c0eb2b7f02752b1fb84a91a0ba6d2f37daa407#stability-of-virtual-environment). If this is the issue, relative benchmarking _within the same virtual environment_ could be a workaround.

PR #141, opened 2023-03-08
fixed depreciated numpy types 

PhilipOsterholz:
I replaced all np.bool, np.int, np.float with bool, int, float respectively, as former are depreciated in numpy 
 1.20.0. The numpy types were just aliases for the builtin types according to: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


hmenke (approved):

hmenke:
Thank you for your contribution!

PR #144, opened 2023-03-26
WIP: Consolidate pireal and picomplex into a single library

hmenke:


doc/sphinx/examples_python/pair_potential_efield_sensitivity.py
```diff
@@ -47,7 +47,7 @@ def getSystems(distance, system_one, fieldshift):
     system_two.diagonalize(1e-3)
 
     # Restrict the calculated eigenenergies
-    system_two.restrictEnergy(fieldshift - 0.015, fieldshift + 0.015)
+    system_two.restrictEnergy(fieldshift.real - 0.015, fieldshift.real + 0.015)
```
seweber, line 50: Why is `.real` required here and in other files? Isn't `SystemOneReal` a specialization for real Hamiltonians?

hmenke: Very good point. I need to investigate.

hmenke: Turns out this was superfluous. I guess I just added a cautionary `.real` after it failed at some intermediate stage when everything turned into `System...Complex` after I removed the `#ifdefs`.

seweber:
> Suggestions for `pireal` (optionally without the pi prefix):

Alternatively, we could allow for `from pairinteraction import SystemOneReal`.

hmenke:
> Alternatively, we could allow for `from pairinteraction import SystemOneReal`.

Then we need `__init__.py` or something like that. I'm not super familiar with Python module loading.

seweber:
> Then we need __init__.py or something like that. I'm not super familiar with Python module loading.

Yes, if `pireal` gets renamed to `newname` so that SWIG produces the file `pairinteraction/newname.py`, we could add the following code to  `pairinteraction/__init__.py`:

```python
from pairinteraction.pairinteraction import *
```

Concerning the name: If you can control the name of the shared library and the Python file generated by SWIG independently, I would prefer having `pairinteraction/binding.py` and `libpairinteraction.so`. If the names should be the same, I would go for `pairinteraction/pairinteraction.py` and `libpairinteraction.so`.

seweber:
Maybe, `libpireal.so` --> `libparinteraction.so`, `_pireal.so` --> `_binding.so` and `pireal.py` --> `binding.py`?

seweber:
The renaming commits could be listed in `.git-blame-ignore-revs`.

seweber (approved):

Issue #133, opened 2023-02-09
MatrixElements::precalculate is called with empty state


hmenke:
The Flatpak builder enables some extra assertions when compiling C++ code which has already uncovered quite a few bugs. When clicking around in the GUI, I found another bug when enabling a magnetic field in z-direction.

Here is the configuration:

<details>

<summary><code>conf.json</code></summary>

[Download `conf.json`](https://github.com/pairinteraction/pairinteraction/files/10696518/conf.json.txt)

```json
{
    "dd": true,
    "deltaESingle": 40.0,
    "deltaJSingle": -1,
    "deltaLSingle": 3,
    "deltaMSingle": -1,
    "deltaNSingle": 3,
    "diamagnetism": true,
    "dq": true,
    "j1": 0.5,
    "j2": 1.5,
    "l1": 0,
    "l2": 1,
    "m1": 0.5,
    "m2": 0.5,
    "maxBx": 0.0,
    "maxBy": 0.0,
    "maxBz": 200.0,
    "maxEx": 0.0,
    "maxEy": 0.0,
    "maxEz": 0.0,
    "minBx": 0.0,
    "minBy": 0.0,
    "minBz": 0.0,
    "minEx": 0.0,
    "minEy": 0.0,
    "minEz": 0.0,
    "missingCalc": true,
    "missingWhittaker": false,
    "n1": 80,
    "n2": 79,
    "precision": 1e-12,
    "qq": true,
    "samebasis": true,
    "species1": "Rb",
    "species2": "Rb",
    "steps": 50
}
```

</details>

<details>

<summary>GDB backtrace</summary>

[Download full backtrace with stack and line information](https://github.com/pairinteraction/pairinteraction/files/10696509/gdb.txt)

```cpp
Thread 2.1 (Thread 0x7f0448848e00 (LWP 2707) "pairinteraction"):
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f0449551537 in __GI_abort () at abort.c:79
[...]

PR #145, opened 2023-03-31
Gui option to use python API

johannes-moegerle:
This is implementing a checkbox to the gui (at Misc) , for using the newer python api.

It still needs a lot of testing as well as benchmarking.

One open bug is occurring when using multiple processes and picomplex (nonzero fields in y direction).


pairinteraction_gui/pairinteraction/app.py
```diff
@@ -3025,13 +3049,28 @@ def startCalc(self):
...
 
-                self.proc = subprocess.Popen(
-                    [path_cpp, "-c", self.path_config, "-o", self.path_cache],
-                    stdout=subprocess.PIPE,
-                    stderr=subprocess.STDOUT,
-                    cwd=self.path_workingdir,
-                    env=dict(os.environ, **ompthreads),
-                )
+                if self.ui.checkbox_use_python_api.isChecked():
+                    self.proc = subprocess.Popen(
```
hmenke, line 3053: I thought the whole point of migrating to the Python interface was that we don't have to fork out a separate process anymore.

hmenke (changes requested):
Also there are zero tests.

pairinteraction_gui/pairinteraction/app.py
seweber: I agree that avoiding forking out a separate process would be nice. However, this does not need to be an immediate goal. For me, the main motivation for getting rid of the old code is increasing maintainability and making it possible to integrate features in the GUI that are currently only available in the Python interface (e.g., supporting strontium). Johanne's approach has the benefit of being a drop-in replacement for the old backend that does not require lots of changes to the existing code. Thus, the risk of breaking existing functionality is reduced. In addition, as it turned out that migrating to the new Python interface is a rather convoluted task, I like the idea of splitting it into several smaller pull requests (e.g. first incorporating Johanne's changes, second avoiding forking out a separate process and maybe rewriting some code in C++, third adding strontium to the GUI, etc.) - especially as this would allow us to get direct user feedback on the current development status.

seweber:
> Also there are zero tests.

It's tricky to have automated tests for the GUI, isn't it? However, if you can look into it, this could be really helpful. Ideally, we would have a test suite that takes the status quo as a reference and ensures that the new Python backend is not breaking anything. But I am not sure whether this would be worth the effort or if we should test the GUI manually.

hmenke:
> It's tricky to have automated tests for the GUI, isn't it?

No, in fact, we do already have a basic GUI test: https://github.com/pairinteraction/pairinteraction/compare/d1c40fc...f99a77e

seweber:
Perfect, maybe, we could use e.g. the Matlab export of the GUI to store results as a reference. Perhaps, we cannot compare the values literally as small deviations are to be expected (e.g. the number of states might be slightly different due to slightly different approaches), but the relevant quantities should not change. For example, we could extract the energies, that correspond to eigenvectors that have a significant overlap with the state of interest, and compare them.

johannes-moegerle:
> Also there are zero tests.

There are now some rudimentary tests checking the energies in the middle of the spectrum against some reference data (for using the old cpp code and the python api).
If you have other things in mind, that definitely should be tested please let me know.

hmenke:
Now that I have fixed #133 you can re-add the test which failed in https://github.com/johannes-moegerle/pairinteraction/commit/f2377e67c64e3f884676ba464ba4abe8079b94ae

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -2578,486 +2594,505 @@ def forbidSamebasis(self):
...
+
     @QtCore.pyqtSlot()
     def validateCores(self):
         sender = self.sender()
         if sender.value() != -1 and sender.value() < 2:
             sender.setValue(2)
 
-    @QtCore.pyqtSlot()
-    def validateQuantumnumbers(self):
+    def validateAllQuantumnumbers(self):
```
seweber, line 2620: The check allows for "Sr3, l=0, and j=0", although it must be j=1 if l=0 for the triplet state.

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -1081,1226 +1111,1200 @@ def getConstDistance(self):
...
+            # --- draw labels at the beginning of the plotting ---
+            if self.ui.groupbox_plot_labels.isChecked() and filestep == 0:
+
+                # probability to find a label inside a basis element
+                if not hasattr(self, "labelprob_energy") or self.labelprob_energy is None:
+                    # nBasis, nLabels # TODO !!! tocsr
+                    self.labelprob = (probs.T * self.labelmat[idx]).tocsr()
+                    self.labelprob_energy = [energies]
+                else:
+                    csr_vappend(self.labelprob, (probs.T * self.labelmat[idx]).tocsr())
```
seweber, line 1683: If different symmetry sectors have a different number of basis elements, the dimension of `self.labelstates` and thus `self.labelmat` are different for the different symmetry sectors. Then, `csr_vappend` produces an invalid `self.labelprob` object that leads to a segfault lateron in the code.

seweber:
I wonder whether everything works correctly with the symmetry sectors:

If I am selecting 2x Rb, n=0, l=0.5, m=0.5 as the initial state, the old and the new backend seem to behave differently. For the new backend, twice as many matrices get diagonalized. Do multiple blocks of the matrix get diagonalized, instead of just diagonalizing the block that transforms under the symmetries in the same way as the state of interest?

pairinteraction_gui/pairinteraction/app.py
johannes-moegerle: we are now using a qthread to run the simulations with the new python backend

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -2999,7 +2999,9 @@ def startCalc(self):
                 params = {k: self.systemdict[k].toUU().magnitude for k in keys}
                 params["button_id"] = button_id
 
-                self.numprocessors = self.systemdict["cores"].magnitude
+                nprc = self.systemdict["cores"].magnitude
+                # set default to number of cores - 1 (leave one core for the GUI/plotting)
+                self.numprocessors = os.cpu_count() - 1 if nprc == 0 else nprc
```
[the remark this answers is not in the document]
johannes-moegerle: i have undone this change, since if the users want to optimize the performance, they should set this anyway by hand

pairinteraction_gui/pairinteraction/app.py
johannes-moegerle: fixed by setting labelprob,... as dict with one entry per symmetry block, now plotting labels for each symmetry block separately

johannes-moegerle:
> I wonder whether everything works correctly with the symmetry sectors:
> 
> If I am selecting 2x Rb, n=0, l=0.5, m=0.5 as the initial state, the old and the new backend seem to behave differently. For the new backend, twice as many matrices get diagonalized. Do multiple blocks of the matrix get diagonalized, instead of just diagonalizing the block that transforms under the symmetries in the same way as the state of interest?

Yes, in fact the symmetry sectors behaviour was not correct/fully implemented before, should be fixed now

seweber (approved):

PR #147, opened 2023-04-05
Enable microarchitecture optimizations and LAPACK support

hmenke:
This enables:

- [x] AVX2 and FMA instructions for compiled binaries. Needs a CPU from at least around 2014.
- [x] Linking to MKL or BLAS/LAPACK (depending on what's available) and offloading all Eigen linear algebra to the linked library.


.github/workflows/macos.yml
```diff
@@ -17,7 +17,7 @@ env:
   CMAKE_BUILD_PARALLEL_LEVEL: 3
   CTEST_PARALLEL_LEVEL: 3
   CTEST_OUTPUT_ON_FAILURE: 1
-  CXXFLAGS: "-Wall -Wextra -Wpedantic"
+  CXXFLAGS: "-Wall -Wextra -Wpedantic -march=broadwell"
```
hmenke, line 20: Not only does this not work in the CI, but will probably also break compatibility with M1. Better leave it out completely rather than trying to find a workaround.

seweber: I agree. Another question: shall we set `WITH_LTO` to `ON` as default in the main `CMakeLists.txt`?

hmenke: In my experience `WITH_LTO` just inflates the build times by something like an order of magnitude without any large gains in performance.

pairinteraction/CMakeLists.txt
```diff
@@ -113,10 +113,37 @@ set_target_properties(wignerSymbols PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_link_libraries(pireal PRIVATE wignerSymbols)
 target_link_libraries(picomplex PRIVATE wignerSymbols)
 
-# Include Eigen
+find_package(BLAS)
```
seweber, line 116: Would it make sense to also set `EIGEN_USE_BLAS`, or is it already included with `EIGEN_USE_LAPACKE_STRICT`?

hmenke: `EIGEN_USE_BLAS` and `EIGEN_USE_LAPACKE_STRICT` are orthogonal. You should benchmark to see whether there is any speedup.

pairinteraction/CMakeLists.txt
```diff
@@ -155,14 +182,35 @@ if(WITH_GSL)
...
-if (MKL_FOUND)
-  target_include_directories(pireal SYSTEM PUBLIC ${MKL_INCLUDE_DIR})
-  target_include_directories(picomplex SYSTEM PUBLIC ${MKL_INCLUDE_DIR})
-  target_link_libraries(pireal PUBLIC ${MKL_LIBRARY})
-  target_link_libraries(picomplex PUBLIC ${MKL_LIBRARY})
+set(MKL_ARCH "intel64" CACHE STRING "MKL architecture")
+set(MKL_LINK "dynamic" CACHE STRING "MKL linkage")
+set(MKL_THREADING "sequential" CACHE STRING "MKL threading backend")
+set(MKL_INTERFACE "lp64" CACHE STRING "MKL interface")
+find_package(MKL CONFIG)
```
seweber, line 189: Could we use MKL instead of OpenBLAS for Eigen if MKL is found (https://eigen.tuxfamily.org/dox/TopicUsingIntelMKL.html)? Maybe, one could gain some additional speedup [[1]](https://www.intel.com/content/www/us/en/developer/articles/technical/performance-comparison-of-openblas-and-intel-math-kernel-library-in-r.html).

hmenke: That's being done already (`-DBLA_VENDOR=Intel10_64lp_seq`) but only on the Debian build which is the only one that has MKL available and isn't deployed anywhere.

seweber: Perfect (`EIGEN_USE_MKL_VML` is not set but does presumably not lead to additional performance gains because we do not make heavy use of the Eigen functions that would be affected by this flag, do you agree? ). It would be great if we could add a remark to the documentation so that advanced users can directly see how to substitute OpenBLAS by MKL, ideally, together with some benchmarking results. However, I see that this is a lot of additional effort. We could keep it in mind and do it once we clean up the installation instructions.

hmenke: I just pushed a commit that takes care of this. If MKL is found, it is used automatically with `EIGEN_USE_MKL_ALL`, otherwise LAPACK is searched normally and `EIGEN_USE_BLAS` and `EIGEN_USE_LAPACKE` are used.

pairinteraction/CMakeLists.txt
seweber: A `run_benchmarks` target, as you have set up in #139, would be handy.  Can you add it to this pull request (just the target, without the CI integration)? Or to the master and then rebase the pull request?

pairinteraction/CMakeLists.txt
```diff
@@ -113,10 +113,77 @@ set_target_properties(wignerSymbols PROPERTIES POSITION_INDEPENDENT_CODE ON)
...
+  if(MKL_FOUND)
+    target_link_libraries(pireal PUBLIC PkgConfig::MKL)
+    target_link_libraries(picomplex PUBLIC PkgConfig::MKL)
+    target_compile_definitions(pireal PUBLIC WITH_INTEL_MKL)
+    target_compile_definitions(picomplex PUBLIC WITH_INTEL_MKL)
+  endif()
+endif()
+
+find_package(BLAS)
+find_package(LAPACK)
```
seweber, line 148: If I only install `libopenblas-dev` and not `liblapacke-dev`, this line still finds lapack: "-- Found LAPACK: /usr/lib/x86_64-linux-gnu/libopenblas.so". However, then, the program is of course not compiling properly. I used cmake version 3.22.1 with `cmake .. -DBLA_VENDOR=OpenBLAS`. Is there an option that we can use to prevent this line from finding lapack if only blas is installed?

hmenke: The problem is that some distributions do not bundle LAPACK**E** (note the trailing **E**, very important) as part of LAPACK (without E). CMake does not offer any bundled detection of LAPACK**E**, which is why there is the `FIXME` further down in the code. Having something that handles all the cases (bundled vs. split package, cross-platform aspects) is not easy. However, it might be a good idea to make it fail when LAPACK**E** is not available. I'll add something for that.

.github/workflows/linux.yml
```diff
@@ -12,9 +12,11 @@ env:
   CMAKE_BUILD_PARALLEL_LEVEL: 2
   CTEST_PARALLEL_LEVEL: 2
   CTEST_OUTPUT_ON_FAILURE: 1
-  CXXFLAGS: "-Wall -Wextra -Wpedantic"
+  CXXFLAGS: "-Wall -Wextra -Wpedantic -march=broadwell"
   QT_QPA_PLATFORM: "offscreen"
   QT_DEBUG_PLUGINS: 1
+  OPENBLAS_NUM_THREADS: 1
+  MKL_NUM_THREADS: 1
```
seweber, line 19: Is this needed with MKL_THREADING sequential?

hmenke: Probably not, but let's play it safe.

pairinteraction/CMakeLists.txt
```diff
@@ -113,10 +113,88 @@ set_target_properties(wignerSymbols PROPERTIES POSITION_INDEPENDENT_CODE ON)
...
+  endif()
+  if(MKL_FOUND)
+    target_link_libraries(pireal PUBLIC PkgConfig::MKL)
+    target_link_libraries(picomplex PUBLIC PkgConfig::MKL)
+    target_compile_definitions(pireal PUBLIC WITH_INTEL_MKL)
+    target_compile_definitions(picomplex PUBLIC WITH_INTEL_MKL)
+  endif()
+endif()
+
+find_package(BLAS)
```
seweber, line 147: If mkl is found, the result of `find_package(BLAS/LAPACK)` will be meaningless - even if it did not return that the packages are found, we would still use intel's blas and lapacke. Thus, I would only search for the packages if mkl is not found.

.github/workflows/linux.yml
```diff
@@ -29,7 +31,9 @@ jobs:
     - name: Fixup Run actions/checkout
       run: git config --global --add safe.directory '*'
     - name: Configure
-      run: cmake -S . -B build/ -DBUILD_TESTING=On -DWITH_COVERAGE=On -DWITH_JULIA=On -DWITH_DOCS=On
+      run: |
+        . /opt/intel/oneapi/setvars.sh
+        cmake -S . -B build/ -DBLA_VENDOR=Intel10_64lp_seq -DBUILD_TESTING=On -DWITH_COVERAGE=On -DWITH_JULIA=On -DWITH_DOCS=On
```
seweber, line 36: Is `DBLA_VENDOR=Intel10_64lp_seq` really needed if MKL is used?

pairinteraction/CMakeLists.txt
hmenke: I don't like so many levels of nesting and it takes away in terms of flexibilty. Now I cannot link to different LAPACK implementations for different targets, e.g. for the benchmarks in #139 where I would like one benchmark that links to OpenBLAS and another that links to MKL.

seweber: I don't see how it takes away flexibility. If MKL is found, the results of find_package(BLAS/LAPACK) were never used anyway if I understood the code correctly. The flexibility to link against OpenBLAS or MKL did not exist because of `if(WITH_LAPACK AND LAPACK_FOUND AND NOT MKL_FOUND)`. However,  I agree that, in principle, one could write the code in such a way that one has this flexibility and that this would be a better solution than my proposal.

seweber:
I did some benchmarking. On my computer, using OpenBlas and march=broadwell halved the time required to diagonalize a matrix of size 2284x2284 :+1:

I also tried using oneapi MKL version 2023.1.0; however, the blas code within MKL failed to compile. Neither gcc version 11.3.0 nor icc version 2021.9.0  worked. Did you encounter similar issues, or do you have a tip to circumvent the problem?

**Error message with gcc:**
```bash
In file included from /opt/intel/oneapi/mkl/2023.1.0/include/mkl_direct_call.h:161,
                 from /opt/intel/oneapi/mkl/2023.1.0/include/mkl.h:47,
                 from /usr/include/eigen3/Eigen/src/Core/util/MKL_support.h:62,
                 from /usr/include/eigen3/Eigen/Core:54,
                 from /usr/include/eigen3/Eigen/Dense:1,
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/dtypes.hpp:23,
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.hpp:23,
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp:20:
/opt/intel/oneapi/mkl/2023.1.0/include/mkl_direct_blas.h: In function ‘void mkl_dc_cgemm(const char*, const char*, const int*, const int*, const int*, const std::complex<float>*, const std::complex<float>*, const int*, const std::complex<float>*, const int*, const std::complex<float>*, std::complex<float>*, const int*)’:
/opt/intel/oneapi/mkl/2023.1.0/include/mkl_direct_blas.h:220:35: error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘double’ to binary ‘operator==’
  220 |         if (m <= 0 || n <= 0 || ((MKL_DC_IS_ZERO(alpha) || k <= 0) && MKL_DC_IS_ONE(beta)))
      |                                   ^~~~~~~~~~~~~~
      |                                   |
      |                                   <unresolved overloaded function type>
      |                                   double
```

**Error message with icc:**
```bash
In file included from /opt/intel/oneapi/mkl/2023.1.0/include/mkl_direct_call.h(162),
                 from /opt/intel/oneapi/mkl/2023.1.0/include/mkl.h(47),
                 from /usr/include/eigen3/Eigen/src/Core/util/MKL_support.h(62),
                 from /usr/include/eigen3/Eigen/Core(54),
                 from /usr/include/eigen3/Eigen/Dense(1),
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/dtypes.hpp(23),
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.hpp(23),
                 from /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp(20):
/opt/intel/oneapi/mkl/2023.1.0/include/mkl_direct_blas.h(220): error: a pointer to a bound function may only be used to call the function
  	if (m <= 0 || n <= 0 || ((MKL_DC_IS_ZERO(alpha) || k <= 0) && MKL_DC_IS_ONE(beta)))
```

hmenke:
The debian job compiles with MKL where it works and the tests pass. What operating system are you using? (`lsb_release -d`)

pairinteraction/CMakeLists.txt
seweber: We could check whether `BLA_VENDOR` is set, and if yes, take the lapacke found by `find_package(BLAS/LAPACK)` instead of intel MKL's lapacke. If you agree that this would be a good solution, I can also look into it.

hmenke:
Could you please configure with `-DCMAKE_EXPORT_COMILE_COMMANDS=On` and send me the output of `grep GreenTensor.cpp build/compile_commands.json`?

seweber:
Thanks for looking into it! Here's the output:

```bash
  "command": "/usr/bin/g++ -DBOOST_ALL_NO_LIB -DBOOST_SERIALIZATION_DYN_LINK -DEIGEN_USE_MKL_ALL -DFMT_HEADER_ONLY=1 -DMKL_Complex16=\"std::complex<double>\" -DMKL_Complex8=\"std::complex<float>\" -DWITH_GSL -DWITH_INTEL_MKL -Dpireal_EXPORTS -I/opt/repos/pairinteraction-hmenke/pairinteraction -I/opt/repos/pairinteraction-hmenke/build/pairinteraction -isystem /opt/repos/pairinteraction-hmenke/build/_deps/wignersymbols-src/include -isystem /opt/intel/oneapi/mkl/2023.1.0/include -isystem /usr/include/eigen3 -Wall -Wextra -Wpedantic -march=broadwell -O2 -g -DNDEBUG -fPIC -fopenmp -o CMakeFiles/pireal.dir/GreenTensor.cpp.o -c /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp",
  "file": "/opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp"
  "command": "/usr/bin/g++ -DBOOST_ALL_NO_LIB -DBOOST_SERIALIZATION_DYN_LINK -DEIGEN_USE_MKL_ALL -DFMT_HEADER_ONLY=1 -DMKL_Complex16=\"std::complex<double>\" -DMKL_Complex8=\"std::complex<float>\" -DUSE_COMPLEX -DWITH_GSL -DWITH_INTEL_MKL -Dpicomplex_EXPORTS -I/opt/repos/pairinteraction-hmenke/pairinteraction -I/opt/repos/pairinteraction-hmenke/build/pairinteraction -isystem /opt/repos/pairinteraction-hmenke/build/_deps/wignersymbols-src/include -isystem /opt/intel/oneapi/mkl/2023.1.0/include -isystem /usr/include/eigen3 -Wall -Wextra -Wpedantic -march=broadwell -O2 -g -DNDEBUG -fPIC -fopenmp -o CMakeFiles/picomplex.dir/GreenTensor.cpp.o -c /opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp",
  "file": "/opt/repos/pairinteraction-hmenke/pairinteraction/GreenTensor.cpp"
```

hmenke:
I'm not entirely sure yet why this fails to compile, but I can tell you why it only fails for you. As you can see the compiler call does not include anything like `-std=c++17`, because probably CMake somehow knows that [GCC 11 defaults to C++17](https://gcc.gnu.org/gcc-11/changes.html), but that is *with* GNU extensions and therefore the `__STRICT_ANSI__` macro is not defined which leads to a wrong (as far as we are concerned) preprocessor branch in `mkl_direct_call.h`
```console
$ echo | /usr/bin/g++ -dM -E -x c++ - | grep STRICT_ANSI
$ echo | /usr/bin/g++ -dM -E -x c++ -std=c++17 - | grep STRICT_ANSI
#define __STRICT_ANSI__ 1
```

seweber:
Great that you have figured it out. I can confirm that adding `-std=c++17`  to `CXXFLAGS` makes the code compile (I also run the benchmark; on my system with an AMD Ryzen 7 5700G the time is very similar to OpenBlas). How should we handle this bug? Can we fix it? Or would the best solution be to add a remark in the documentation that, for compilation, it could be required to set this flag manually?

hmenke:
I found the source of the bug. It is ` -DMKL_Complex8="std::complex<float>" -DMKL_Complex16="std::complex<double>"`. That fucks up some of the MKL internals.

hmenke:
Turns out, this is a probably also a CMake bug, since it silently drops `CXX_EXTENSIONS OFF`: https://gitlab.kitware.com/cmake/cmake/-/issues/22224

I don't think it's worth working around bugs in some very specific configurations.

hmenke:
Poking around a bit, I found that Ubuntu 22.04 is shipping quite a lot of broken stuff. Workarounds for all of that are on `master` now, so you should now be able to build without any errors.

seweber:
Thanks. Now it seems to work except for warnings `Warning 202: Could not evaluate expression '__has_include (<boost/serialization/version.hpp>)'` that I obtain using gcc.

seweber:
> We could check whether `BLA_VENDOR` is set, and if yes, take the lapacke found by `find_package(BLAS/LAPACK)` instead of intel MKL's lapacke. If you agree that this would be a good solution, I can also look into it.

I implemented such a mechanism in commit [269d1aa](https://github.com/pairinteraction/pairinteraction/pull/147/commits/269d1aa50f605dc7888eac8cd739a633f3855b79). Using MKL and OpenBLAS is mutually exclusive because the lapack implementation of MKL is taken as soon as we link against MKL.

seweber:
Can we set `OPENBLAS_NUM_THREADS=1` and `MKL_NUM_THREADS=1` before starting the GUI to ensure that the multithreading by OpenBLAS does not interfere with our multithreading?

pairinteraction/CMakeLists.txt
```diff
@@ -113,19 +113,111 @@ set_target_properties(wignerSymbols PROPERTIES POSITION_INDEPENDENT_CODE ON)
...
+  elseif(MKL_THREADING STREQUAL "tbb_thread")
+    pkg_check_modules(MKL mkl-${MKL_LINK}-${MKL_INTERFACE}-tbb IMPORTED_TARGET)
+  endif()
+endif()
+
+# Find a different lapack vendor
+find_package(LAPACK)
+
+# Determine which lapack vendor to use
+if(WITH_LAPACK AND MKL_FOUND AND (NOT BLA_VENDOR OR BLA_VENDOR MATCHES "All" OR (BLA_VENDOR MATCHES "^Intel")))
```
hmenke, line 146: This condition is completely unreadable and surely does not work correctly, because `BLA_VENDOR MATCHES "All"` might also select MKL if there is no other LAPACK implementation. To be quite honest I think the code is now worse that before.

pairinteraction/CMakeLists.txt
```diff
@@ -113,19 +113,111 @@ set_target_properties(wignerSymbols PROPERTIES POSITION_INDEPENDENT_CODE ON)
...
+  # Using Intel MKL from Eigen
+  message(STATUS "Using Intel MKL from Eigen")
+  target_compile_definitions(pireal PUBLIC
+  EIGEN_USE_MKL_ALL
+  "-DMKL_Complex8=std::complex<float>"
+  "-DMKL_Complex16=std::complex<double>")
+  target_compile_definitions(picomplex PUBLIC
+  EIGEN_USE_MKL_ALL
+  "-DMKL_Complex8=std::complex<float>"
+  "-DMKL_Complex16=std::complex<double>")
```
hmenke, line 169: Why was all the indentation removed?

pairinteraction/CMakeLists.txt
seweber: The code did not work at all before. It did not provide any working option to select whether one prefers to use MKL or OpenBLAS. MKL was always used as soon as MKL was installed.

The condition ensures that MKL is used only if a) MKL is found and lapack should be used *and* b) no vendor is specified that does not start with "Intel". `BLA_VENDOR MATCHES "All"` is necessary because cmake sets `BLA_VENDOR` to `All` if it is not specified by the user. At least, this was the behavior of my cmake version.

hmenke:
> Can we set `OPENBLAS_NUM_THREADS=1` and `MKL_NUM_THREADS=1` before starting the GUI to ensure that the multithreading by OpenBLAS does not interfere with our multithreading?

```patch
diff --git a/pairinteraction_gui/pairinteraction/app.py b/pairinteraction_gui/pairinteraction/app.py
index e4a6a192..26de76bb 100644
--- a/pairinteraction_gui/pairinteraction/app.py
+++ b/pairinteraction_gui/pairinteraction/app.py
@@ -3030,7 +3030,7 @@ class MainWindow(QtWidgets.QMainWindow):
                     stdout=subprocess.PIPE,
                     stderr=subprocess.STDOUT,
                     cwd=self.path_workingdir,
-                    env=dict(os.environ, **ompthreads),
+                    env=dict(os.environ, **ompthreads, **{"OPENBLAS_NUM_THREADS": 1, "MKL_NUM_THREADS": 1}),
                 )
 
                 self.starttime = time()
```

hmenke:
> Thanks. Now it seems to work except for warnings `Warning 202: Could not evaluate expression '__has_include (<boost/serialization/version.hpp>)'` that I obtain using gcc.

This warning comes from SWIG and can be safely ignored.

pairinteraction_gui/pairinteraction/app.py
```diff
@@ -3030,7 +3030,7 @@ def startCalc(self):
                     stdout=subprocess.PIPE,
                     stderr=subprocess.STDOUT,
                     cwd=self.path_workingdir,
-                    env=dict(os.environ, **ompthreads),
+                    env=dict(os.environ, **ompthreads, **{"OPENBLAS_NUM_THREADS": 1, "MKL_NUM_THREADS": 1}),
```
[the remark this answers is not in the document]
hmenke: The fix is simple:

```suggestion
                    env=dict(os.environ, **ompthreads, **{"OPENBLAS_NUM_THREADS": "1", "MKL_NUM_THREADS": "1"}),
```

pairinteraction/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ target_link_libraries(picomplex PUBLIC Eigen3::Eigen)
 # Find MKL
 set(MKL_ARCH "intel64" CACHE STRING "MKL architecture")
 set(MKL_LINK "dynamic" CACHE STRING "MKL linkage")
-set(MKL_THREADING "sequential" CACHE STRING "MKL threading backend")
+set(MKL_THREADING "intel_thread" CACHE STRING "MKL threading backend")
```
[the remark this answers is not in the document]
seweber: In the long run, it would be cool if we could make the optimizations work with Windows as well. `gnu_thread` would not work there, I guess. Is there an easy way to enable threading cross-platform? If not, I am fine with `gnu_thread`.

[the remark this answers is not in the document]
seweber: I just looked into `MKLConfig.cmake`: If no threading option is selected, it tries to find a suitable multithreading option by itself. Maybe, it is best to remove `set(MKL_THREADING "sequential" CACHE STRING "MKL threading backend")` and rely on this mechanism?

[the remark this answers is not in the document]
seweber: Hmm, neither removing this line nor using `gnu_thread` worked with the CI.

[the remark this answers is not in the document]
hmenke: I found the problem, but I don't have a solution yet. In the build log you see
```
-- Found MKL: /opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin/libiomp5.so  
```
but that file is a symbolic link and the dynamic linker seems to not like that.

For `gnu_thread` the problem is a bug (or rather inadequacy) of `MKLConfig.cmake`. It inspects `CMAKE_C_COMPILER_ID` if that is `GNU`, but ignores `CMAKE_CXX_COMPILER`. So the simplest solution for that would probably be
```cmake
set(CMAKE_C_COMPILER_ID "${CMAKE_CXX_COMPILER_ID}")
set(MKL_THREADING "gnu_thread" CACHE STRING "MKL threading backend")
```

.github/workflows/linux.yml
```diff
@@ -30,17 +30,18 @@ jobs:
         submodules: true
     - name: Fixup Run actions/checkout
       run: git config --global --add safe.directory '*'
+    - name: Setup oneapi environment
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        printenv >> $GITHUB_ENV
```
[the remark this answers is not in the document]
hmenke: The real problem is that `/opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin/` (the location of `libiomp5.so`) is missing from RPATH.

```
$ readelf -d build/pairinteraction/libpicomplex.so

Dynamic section at offset 0x7d4670 contains 38 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libsqlite3.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libgsl.so.25]
 0x0000000000000001 (NEEDED)             Shared library: [libgomp.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_lp64.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_thread.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_core.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libiomp5.so]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000e (SONAME)             Library soname: [libpicomplex.so]
 0x000000000000001d (RUNPATH)            Library runpath: [/opt/intel/oneapi/mkl/2023.1.0/lib/intel64]
 0x000000000000000c (INIT)               0x4aa000
 0x000000000000000d (FINI)               0x6f4d00
 0x0000000000000019 (INIT_ARRAY)         0x7d4428
 0x000000000000001b (INIT_ARRAYSZ)       296 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x7d4550
 0x000000000000001c (FINI_ARRAYSZ)       152 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x298
 0x0000000000000005 (STRTAB)             0x8d660
 0x0000000000000006 (SYMTAB)             0x253f8
 0x000000000000000a (STRSZ)              2188688 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x7d6000
 0x0000000000000002 (PLTRELSZ)           335280 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x458030
 0x0000000000000007 (RELA)               0x2ac8c0
 0x0000000000000008 (RELASZ)             1750896 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0x2ac6d0
 0x000000006fffffff (VERNEEDNUM)         7
 0x000000006ffffff0 (VERSYM)             0x2a3bf0
 0x000000006ffffff9 (RELACOUNT)          72308
 0x0000000000000000 (NULL)               0x0
```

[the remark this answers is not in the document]
seweber: > Please don't do this. Making whether it works dependent on the presence of random environment variables is not good and will just cause confusion.

To my understanding, oneapi is meant to be used in an environment where the environment variables are set with `. /opt/intel/oneapi/setvars.sh`. I am not sure whether this is the best solution and whether one can use it to create packages that are portable to other systems. Still, it seems to work in commit [d77438d](https://github.com/pairinteraction/pairinteraction/pull/147/commits/d77438d5123e5b71f5d6a81469b44decdb643a50). For me, it would be fine to use this solution and think about providing packages that make use of mkl at a later point. Then, for the moment, linking against mkl would be an option for advanced users that compile pairinteraction on their own.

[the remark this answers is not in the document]
hmenke: I'd say, revert all of that and go with `MKL_THREADING="sequential"` for now until we find an *actual* solution.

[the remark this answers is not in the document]
seweber: I wonder whether there is a nice solution. Manually modifying the RPATH would also be hacky. If I search for the error or how oneapi is used in other github actions, the standard answer seems to be to run `. /opt/intel/oneapi/setvars.sh`. For example, [pybind11](https://github.com/pybind/pybind11/blob/da91926295638b2b1c0c85a568a2d66ca3646565/.github/workflows/ci.yml) is running it multiple times.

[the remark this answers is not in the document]
hmenke: > To my understanding, oneapi is meant to be used in an environment where the environment variables are set

I would agree if we were talking about a system that pairinteraction was *not* compiled on, but this is not the case here. The point is that `MKLConfig.cmake` is buggy and needs to be fixed.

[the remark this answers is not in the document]
seweber: Hmm, but this is nothing we can fix easily, isn't it? Then, I would prefer to have the `setvars.sh` workaround. I find it handy to have the possibility to use the multithreaded version of mkl, especially as I am currently looking into benchmarking different eigensolver implementations.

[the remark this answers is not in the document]
hmenke: I fixed it already, see https://github.com/pairinteraction/pairinteraction/pull/147/commits/73251d8657cbb08d443aac4362b876ff86bd7058 and I will submit a bug report upstream once I find out where to do that. `MKLConfig.cmake` sets a variable `OMP_LIBRARY` which contains the full path to `libiomp5`, so all that's left is to extract the directory, dereference all the symlinks, and add it to RPATH.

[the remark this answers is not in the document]
seweber: Oh, I just saw it. I am fine with this solution (except for the way `set(MKL_THREADING "intel_thread" CACHE STRING "MKL threading backend")` is called - as written previously, I would prefer to let MKLConfig.cmake determine a threading backend whenever possible). However, somehow, your force push destroyed some bug fixes that I did to other files.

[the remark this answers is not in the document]
hmenke: I just squashed some of these fixes into https://github.com/pairinteraction/pairinteraction/pull/147/commits/6f19346af3be2a3be3cfd23f89d71d3400ecd259

About https://github.com/pairinteraction/pairinteraction/commit/65c14b08a684f4d0d93159e06e1a8cb9b264da1b I'm not so sure. Seems a bit excessive to set these environment variables for all tests. This might shadow problems arising from multithreaded LAPACK. Maybe only for the `python.parallelization` test would suffice, or just set it from within the test. A call to `os.environ` should then precede `import pairinteraction`.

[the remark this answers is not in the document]
seweber: Ok, I completely agree that https://github.com/pairinteraction/pairinteraction/commit/65c14b08a684f4d0d93159e06e1a8cb9b264da1b is a bit excessive. The cleanest solution would be to neither set these environment variables with `set_tests_properties` nor within the `.yml` files but for each individual test. However, I don't know how straightforward it is to figure out which tests require these environment variables and which do not. Without `MKL_NUM_THREADS=1`, surprisingly many tests failed. I just added a `FIXME` to the code as a reminder, see https://github.com/pairinteraction/pairinteraction/pull/147/commits/14d74feb57dcfa753ba16e67fa974ca23e097b39. For me, this would be fine for the moment, but feel free to investigate it in more detail.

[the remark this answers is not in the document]
seweber: Strangely, calculated eigenenergies, etc. are different if multiple threads are used. Maybe, this is just because of a different order of the eigenvalues. But we should look into it before merging the pull request to exclude that a user obtains wrong results if the user does not set the environment variables.

seweber (commented):
Why are calculated eigenenergies, etc. different if multiple threads are used?

hmenke:
There is definitely a race-condition somewhere because the values are different each time.

hmenke:
> It turns out that the incorrect matrix product is a result of
> gomp + iomp library clash: octave is linked against the GNU OMP,
> while libmkl-rt.so invokes Intel(LLVM) OMP by default.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=921207

hmenke:
Also could you please test stuff on your own fork or locally instead of filling up my branch with `fixup!` commits?

seweber:
> Also could you please test stuff on your own fork or locally instead of filling up my branch with `fixup!` commits?

A simple `autosquash` removes the `fixup!` commits, so I have not thought that they are a problem ... but, of course, I can also use a fork to get things working with the CI.

hmenke:
Anyway, `intel_thread` is out. It's too buggy, giving us linking problems and conflicting with gomp. I've determined `gnu_thread` to work, but that needs manual intervention. So back to `sequential`.

seweber (approved):

PR #149, opened 2023-05-05
LAPACK support for other targets

hmenke:
- [x] macOS
- [x] Windows

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,13 +18,34 @@ env:
   CTEST_OUTPUT_ON_FAILURE: 1
   CMAKE_BUILD_TYPE: Release
   CXXFLAGS: "/W3 /Ox /arch:AVX2"
-  QT_QPA_PLATFORM: "offscreen"
-  QT_DEBUG_PLUGINS: 1
 
 jobs:
   build:
     runs-on: windows-2019
     steps:
+      # https://www.intel.com/content/www/us/en/developer/articles/tool/oneapi-standalone-components.html
+      # https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-windows/2023-1/install-with-command-line.html
+      # https://github.com/oneapi-src/oneapi-ci
+      - name: Install Intel oneAPI Math Kernel Library (oneMKL)
+        shell: cmd
+        run: |
+          @echo on
+
+          REM SPDX-FileCopyrightText: 2022 Intel Corporation
+          REM
+          REM SPDX-License-Identifier: MIT
+
+          set URL=https:/registrationcenter-download.intel.com/akdlm/IRC_NAS/c961e083-5685-4f0b-ada5-c6cf16f561dd/w_BaseKit_p_2023.1.0.47256_offline.exe
+          set COMPONENTS=intel.oneapi.win.mkl.devel:intel.oneapi.win.tbb.devel:intel.oneapi.win.dpl
+
+          curl.exe --output %TEMP%\webimage.exe --url %URL% --retry 5 --retry-delay 5
+          start /b /wait %TEMP%\webimage.exe -s -x -f webimage_extracted --log extract.log
+          del %TEMP%\webimage.exe
+          webimage_extracted\bootstrapper.exe -s --action install --components=%COMPONENTS% --eula=accept -p=NEED_VS2017_INTEGRATION=0 -p=NEED_VS2019_INTEGRATION=0 -p=NEED_VS2022_INTEGRATION=0 --log-dir=.
+          set installer_exit_code=%ERRORLEVEL%
+          rd /s/q "webimage_extracted"
+          exit /b %installer_exit_code%
+
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
@@ -49,7 +70,9 @@ jobs:
         shell: cmd
         run: |
           call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          call "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
           cmake -S . -B build/ -G "Visual Studio 16 2019" -DBUILD_TESTING=On ^
+              -DMKL_THREADING=intel_thread ^
               -DPython3_FIND_REGISTRY=LAST ^
               -DCMAKE_BUILD_TYPE=${{ env.CMAKE_BUILD_TYPE }} ^
               -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE ^
@@ -79,6 +102,9 @@ jobs:
           Copy-Item pairinteraction\Release\* pairinteraction\unit_test\Release -Recurse -Force -Verbose
           cmake --build . --target RUN_TESTS --config ${{ env.CMAKE_BUILD_TYPE }}
         working-directory: build/
+        env:
+          QT_QPA_PLATFORM: "offscreen"
+          QT_DEBUG_PLUGINS: 1
 
       - name: Run debugging session in tmate
         if: github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled && always()
--- a/pairinteraction/CMakeLists.txt
+++ b/pairinteraction/CMakeLists.txt
@@ -140,6 +140,48 @@ if (NOT MKL_FOUND AND PKG_CONFIG_FOUND)
   endif()
 endif()
 
+if(MSVC)
+  # https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-windows/2023-1/contents-of-the-redist-intel64-directory.html
+  file(GLOB MKL_DLLS
+    # Threading layer
+    "${MKL_ROOT}/redist/intel64/mkl_${MKL_THREADING}*.dll"
+    # Computational layer
+    "${MKL_ROOT}/redist/intel64/mkl_core*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_def*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_mc*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_avx*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_def*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_mc*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_avx*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_cmpt*.dll"
+    "${MKL_ROOT}/redist/intel64/libimalloc.dll")
+  if(NOT MKL_DLLS)
+    message(WARNING "MKL DLLs could not be located! Dynamic lookup will probably fail.")
+  endif()
+  if(MKL_THREADING STREQUAL "intel_thread")
+    # Implementation detail of MKLConfig.cmake, the pkg-config files do not expose these variables
+    if(EXISTS "${OMP_DLL_DIR}/${OMP_DLLNAME}")
+      list(APPEND MKL_DLLS ${OMP_DLL_DIR}/${OMP_DLLNAME})
+    else()
+      message(WARNING "Intel OpenMP DLL cloud not be located! Dynamic lookup will probably fail.")
+    endif()
+  elseif(MKL_THREADING STREQUAL "tbb_thread")
+    # FIXME: Do we have to add any additional DLLs?
+  endif()
+  set(MKL_DLLS_TO_COPY "${MKL_DLLS}" CACHE STRING "Intel MKL DLLs to copy into the target dir (;-list, ignored if not on Windows)")
+  unset(MKL_DLLS)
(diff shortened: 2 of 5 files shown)


.github/workflows/windows.yml
```diff
@@ -74,11 +96,15 @@ jobs:
 
       - name: Tests
         run: |
+          Copy-Item "C:\Program Files (x86)\Intel\oneAPI\mkl\latest\redist\intel64\*.dll" pairinteraction\Release\ -Force -Verbose
```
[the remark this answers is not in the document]
seweber: Can you use CMake to copy these DLLs? To my understanding, copying these DLLs is always required under Windows and is nothing specific to our GitHub workflow (MKLConfig.cmake does not take care of these DLLs).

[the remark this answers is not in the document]
hmenke: The files inside a folder named `redist` better be redistributable.

I'm not a big fan of copying DLLs inside of CMake, because CMake is a buildsystem generator and not a packaging system. Instead `setup.py bdist_wheel` or the `pyinstaller` should copy those DLLs, because these are the programs that actually need them. Didn't we have to copy MKL DLLs in the past when we used Conda on Appveyor? How did we do it back then?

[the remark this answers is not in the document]
seweber: > These are all the files in the redist directory

From my point of view, we can copy them all except for the files that correspond to feature that we do not enable (the MKLConfig.cmake provides hints which DLLs correspond to functionality that we do not use like scalapack).

seweber:
Sounds good :+1:

pairinteraction/CMakeLists.txt
```diff
@@ -140,6 +140,26 @@ if (NOT MKL_FOUND AND PKG_CONFIG_FOUND)
...
+function(target_link_mkl TARGET)
+  if(MSVC AND MKL_FOUND AND MKL_ROOT)
+    set(MKL_REDIST_DIR ${MKL_ROOT}/redist/intel64)
+    set(MKL_DLLS
+      ${MKL_REDIST_DIR}/mkl_avx.2.dll
+      ${MKL_REDIST_DIR}/mkl_avx2.2.dll
+      ${MKL_REDIST_DIR}/mkl_avx512.2.dll
+      ${MKL_REDIST_DIR}/mkl_core.2.dll
+      ${MKL_REDIST_DIR}/mkl_def.2.dll
+      ${MKL_REDIST_DIR}/mkl_sequential.2.dll)
```
hmenke, line 152: Not sure whether that is really the set of necessary and sufficient libraries, but it works.

hmenke: Maybe also this?
```suggestion
    set(MKL_DLLS
      ${MKL_REDIST_DIR}/mkl_avx.2.dll
      ${MKL_REDIST_DIR}/mkl_avx2.2.dll
      ${MKL_REDIST_DIR}/mkl_avx512.2.dll
      ${MKL_REDIST_DIR}/mkl_core.2.dll
      ${MKL_REDIST_DIR}/mkl_def.2.dll
      ${MKL_REDIST_DIR}/mkl_${MKL_THREADING}.2.dll)
```

pairinteraction/CMakeLists.txt
```diff
@@ -140,6 +140,31 @@ if (NOT MKL_FOUND AND PKG_CONFIG_FOUND)
...
+function(target_link_mkl TARGET)
+  if(MSVC AND TARGET MKL::MKL)
+    set(MKL_REDIST_DIR ${MKL_ROOT}/redist/intel64)
+    list(APPEND MKL_DLLS
+      ${MKL_REDIST_DIR}/mkl_avx.2.dll
+      ${MKL_REDIST_DIR}/mkl_avx2.2.dll
+      ${MKL_REDIST_DIR}/mkl_avx512.2.dll
+      ${MKL_REDIST_DIR}/mkl_core.2.dll
+      ${MKL_REDIST_DIR}/mkl_def.2.dll
+      ${MKL_REDIST_DIR}/mkl_${MKL_THREADING}.2.dll)
```
seweber, line 153: I think it is better to use `${MKL_REDIST_DIR}/mkl_avx*.dll` etc. as it is done in the MKLConfig.cmake, instead of fixing the version of the DLLs to 2. To always copy the newest version of the DLLs, one could write for the foreach loop (untested):

```cmake
foreach(EXPRESSION ${MKL_DLLS})
  file(GLOB FOUND_FILES EXPRESSION)
  list(SORT FOUND_FILES)
  list(REVERSE FOUND_FILES)
  list(GET FILE 0 FOUND_FILES)
  add_custom_command(TARGET ${TARGET} POST_BUILD
     COMMAND ${CMAKE_COMMAND} -E copy_if_different ${FILE} $<TARGET_FILE_DIR:${TARGET}>)
endforeach()
```

seweber: Should we add the DLLs "mkl_mc", "mkl_mc3", "libimalloc", and maybe the "mkl_vml_..." DLLs to be on the safe side? My fear is that depending on the processor on which pairinteraction runs, MKL wants to load some of these DLLs as well.

seweber (changes requested):

pairinteraction/CMakeLists.txt
hmenke: 1. We don't need to any of `mkl_mc.<version>.dll` or `mkl_mc3.<version>.dll` because these are only needed for processors that do not support AVX but only the older SSE instruction set. However, pairinteraction is now compiled with explicit AVX2 support, so any processor without AVX is unsupported anyway.

2. The `mkl_vml_*.<version>.dll` files are for the [Vector Mathematical Functions](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-1/vector-mathematical-functions.html) but we are not using any of those. This is a separate API.

3. We also do not at this moment try to [make Intel MKL use another memory allocator](https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-windows/2023-1/redefining-memory-functions.html) which is what `libimalloc.dll` is for.

    > Intel® oneAPI Math Kernel Library memory management by default uses standard C run-time memory functions to allocate or free memory. These functions can be replaced using memory renaming.

4. I would really like to avoid globbing because then we have to think about what to do if the glob expression returned something empty. Do we fail? But what if one of the DLLs we're trying to glob has been removed in that particular version, so the glob will always fail? The bottomline is that when we upgrade MKL we have to read the documentation anyway, so there is not really a point in complicating things. Also generally I prefer this failing early and loudly rather than late and silently.

hmenke: BTW, you proposal fails in the presence of versioned and unversioned DLLs because it will always pick the unversioned one this way.
```cmake
#!/usr/bin/env -S cmake -P
list(APPEND FOUND_FILES mkl_def.dll mkl_def.1.dll mkl_def.2.dll) # pretend that we globbed this
list(SORT FOUND_FILES)
list(REVERSE FOUND_FILES)
message(${FOUND_FILES})
list(GET FOUND_FILES 0 FILE)
message(${FILE})
```
prints
```console
$ ./test.cmake
mkl_def.dllmkl_def.2.dllmkl_def.1.dll
mkl_def.dll
```

seweber: > The mkl_vml_*.<version>.dll files are for the [Vector Mathematical Functions](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-1/vector-mathematical-functions.html) but we are not using any of those. This is a separate API.

Isn't this used for some functions in eigen?

hmenke: Ah, yes, you're right, but only for array operations that we currently don't use. But we should indeed ship the DLLs in this case because in principle these functions could sneak in at some point.

| Code example              | MKL routines |
|---------------------------|--------------|
| `v2=v1.array().sin();`    | `v?Sin`      |
| `v2=v1.array().asin();`   | `v?Asin`     |
| `v2=v1.array().cos();`    | `v?Cos`      |
| `v2=v1.array().acos();`   | `v?Acos`     |
| `v2=v1.array().tan();`    | `v?Tan`      |
| `v2=v1.array().exp();`    | `v?Exp`      |
| `v2=v1.array().log();`    | `v?Ln`       |
| `v2=v1.array().sqrt();`   | `v?Sqrt`     |
| `v2=v1.array().square();` | `v?Sqr`      |
| `v2=v1.array().pow(1.5);` | `v?Powx`     |

https://eigen.tuxfamily.org/dox/TopicUsingIntelMKL.html

hmenke: Following this discussion I think just stupidly copying all of the DLLs is the easiest.
```patch
diff --git a/pairinteraction/CMakeLists.txt b/pairinteraction/CMakeLists.txt
index 1c7d777..1d4007d 100644
--- a/pairinteraction/CMakeLists.txt
+++ b/pairinteraction/CMakeLists.txt
@@ -144,13 +144,7 @@ endif()
 function(target_link_mkl TARGET)
   if(MSVC AND TARGET MKL::MKL)
     set(MKL_REDIST_DIR ${MKL_ROOT}/redist/intel64)
-    list(APPEND MKL_DLLS
-      ${MKL_REDIST_DIR}/mkl_avx.2.dll
-      ${MKL_REDIST_DIR}/mkl_avx2.2.dll
-      ${MKL_REDIST_DIR}/mkl_avx512.2.dll
-      ${MKL_REDIST_DIR}/mkl_core.2.dll
-      ${MKL_REDIST_DIR}/mkl_def.2.dll
-      ${MKL_REDIST_DIR}/mkl_${MKL_THREADING}.2.dll)
+    file(GLOB MKL_DLLS "${MKL_REDIST_DIR}/*.dll")
     if(MKL_THREADING STREQUAL "intel_thread")
       # Implementation detail of MKLConfig.cmake, the pkg-config files do not expose these variables
       list(APPEND MKL_DLLS ${OMP_DLL_DIR}/${OMP_DLLNAME})
```

seweber: How much does this increase the file size? If the increase is small, I like this solution as it is perhaps the most robust thing that we can do. If MKL's "cluster libraries" are increasing the file size a lot, I would still prefer globing individual DLLs (https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-windows/2023-1/contents-of-the-redist-intel64-directory.html is using a placeholder for the version in the file names of the DLLs, so my proposed globing should be fine).

hmenke: So you mean we should glob *only* versioned DLLs and ignore unversioned?

hmenke: The cluster libraries are tiny, but the two `mkl_sycl*.dll` are quite large, see the listing in my previous comment: https://github.com/pairinteraction/pairinteraction/pull/149#discussion_r1190748794

pairinteraction/CMakeLists.txt
```diff
@@ -140,6 +140,48 @@ if (NOT MKL_FOUND AND PKG_CONFIG_FOUND)
...
+    # Computational layer
+    "${MKL_ROOT}/redist/intel64/mkl_core*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_def*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_mc*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_avx*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_def*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_mc*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_avx*.dll"
+    "${MKL_ROOT}/redist/intel64/mkl_vml_cmpt*.dll"
+    "${MKL_ROOT}/redist/intel64/libimalloc.dll")
```
hmenke, line 157: [`MKLConfig.cmake`](https://termbin.com/wh1x) (line 634) only looks for versioned DLLs. However, there are some issues with us also looking for only versioned DLLs.

1. MKL looks for only some versioned DLLs, e.g. `mkl_intel_thread.2.dll`, but this library in turn is linked against, say `mkl_avx2.3.dll`. We can't resolve this transitive versioning by only introspecting the targets exposed my `MKLConfig.cmake`.
2. I estimate the chances that there are both versioned and unversioned DLLs in the same redist folder as relatively slim, so extensively globbing should not pose a big problem.

This is what has been globbed for this particular installation of MKL:
```
libimalloc.dll
libiomp5md.dll
mkl_avx.2.dll
mkl_avx2.2.dll
mkl_avx512.2.dll
mkl_core.2.dll
mkl_def.2.dll
mkl_intel_thread.2.dll
mkl_mc.2.dll
mkl_mc3.2.dll
mkl_vml_avx.2.dll
mkl_vml_avx2.2.dll
mkl_vml_avx512.2.dll
mkl_vml_cmpt.2.dll
mkl_vml_def.2.dll
mkl_vml_mc.2.dll
mkl_vml_mc3.2.dll
```

hmenke: On top of that we could also define `MKL_DLLS` as a `CACHE` variable, so it can be overridden by a user on the command line.

seweber (approved):

PR #151, opened 2023-05-08
Speed up diagonalization by calling LAPACKE_dsyevd/zheevd

seweber:


pairinteraction/SystemBase.hpp
```diff
@@ -61,7 +61,10 @@
 #ifdef WITH_INTEL_MKL
 #define MKL_Complex8 std::complex<float>
 #define MKL_Complex16 std::complex<double>
+#include <mkl_lapacke.h>
 #include <mkl_solvers_ee.h>
+#elif defined EIGEN_USE_LAPACKE
+#include <lapacke.h>
```
hmenke, line 67: These interfaces are standardized, so if the header does not exist you could also just hardcode the prototypes. Something like this (untested)

```suggestion
#elif defined EIGEN_USE_LAPACKE
#  if __has_include(<lapacke.h>)
#    include <lapacke.h>
#  else
#    ifndef lapack_int
#      define lapack_int int // lp64
#    endif // lapack_int
extern "C" {
lapack_int LAPACKE_dsyevd( int matrix_layout, char jobz, char uplo, lapack_int n,
                           double* a, lapack_int lda, double* w );
lapack_int LAPACKE_zheevd( int matrix_layout, char jobz, char uplo, lapack_int n,
                           lapack_complex_double* a, lapack_int lda,
                           double* w );
}
#  endif //  __has_include(<lapacke.h>)
```

hmenke: I just saw that both manylinux and opensuse actually have a `lapacke.h` header, but it's located at `/usr/include/openblas/lapacke.h` and this is not automatically found.

hmenke (approved):
Should probably be merged before #149

Issue #161, opened 2023-06-22
"Clear cache" fails to remove "cache_matrix_real_new"


seweber:
With the new Python API enabled, "Clear cache" fails to remove "cache_matrix_real_new" on some Linux systems (sometimes with the error message "Directory not empty", sometimes with the error message "Device or resource busy"). It looks like this is caused by the workers still having a `MatrixElementCache` open. Perhaps, we have to shut down the pool of workers if  "Clear cache" is called and recreate the pool afterward.

Issue #159, opened 2023-06-21
Option "using Whittaker functions" not working with the new Python API


seweber:
If the new Python API is activated in the GUI and the option "using Whittaker functions" selected, an error is raised:

```bash
File "/media/cat/sebastian/.local/lib/python3.10/site-packages/pairinteraction_gui/pipy/atom.py", line 66, in system
    self._createSystem()
File "/media/cat/sebastian/.local/lib/python3.10/site-packages/pairinteraction_gui/pipy/atom.py", line 345, in _createSystem
    system = pi.SystemOne(config.species(), self.getCache())
File "/media/cat/sebastian/.local/lib/python3.10/site-packages/pairinteraction_gui/pipy/atom.py", line 293, in getCache
    pi.MatrixElementCache.setMethod(pi.WHITTAKER)
TypeError: MatrixElementCache.setMethod() missing 1 required positional argument: 'm'
```

PR #162, opened 2023-06-23
Fix issues

johannes-moegerle:
see #159 and #161

### DIFF
--- a/pairinteraction_gui/worker.py
+++ b/pairinteraction_gui/worker.py
@@ -16,6 +16,12 @@
 # along with the pairinteraction GUI. If not, see <http://www.gnu.org/licenses/>.
 from queue import Queue
 
+# FIXME once we assume PyQt5.version >= 5.15, we can remove the except
+# https://www.riverbankcomputing.com/static/Docs/PyQt5/incompatibilities.html#importing-the-sip-module
+try:
+    from PyQt5 import sip
+except ImportError:
+    import sip
 from PyQt5.QtCore import QThread, pyqtSignal
 
 
@@ -29,7 +35,8 @@ class Worker(QThread):
 
     def __del__(self):
         self.exiting = True
-        self.wait()
+        if not sip.isdeleted(self):
+            self.wait()
 
     def execute(self, stdout):
         self.stdout = stdout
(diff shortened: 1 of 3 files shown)


hmenke:
Maybe this could also be added to the patchset. This is to (possibly?) work around the test failure that I observed on Flathub: https://buildbot.flathub.org/#/builders/35/builds/5977/steps/8/logs/stdio
```diff
diff --git a/pairinteraction_gui/worker.py b/pairinteraction_gui/worker.py
index febc284e..b99ba013 100644
--- a/pairinteraction_gui/worker.py
+++ b/pairinteraction_gui/worker.py
@@ -16,6 +16,7 @@
 # along with the pairinteraction GUI. If not, see <http://www.gnu.org/licenses/>.
 from queue import Queue
 
+from PyQt5 import sip
 from PyQt5.QtCore import QThread, pyqtSignal
 
 
@@ -29,7 +30,8 @@ class Worker(QThread):
 
     def __del__(self):
         self.exiting = True
-        self.wait()
+        if not sip.isdeleted(self):
+            self.wait()
 
     def execute(self, stdout):
         self.stdout = stdout
```

johannes-moegerle:
We are also calling QThread.wait in pipy_thready.py (see PipyThread.terminate), as well as a few times directly in app.py.
Should this workaround also be added there?

hmenke:
I think the culprit here is that `self.wait()` is called inside `__del__` and the Qt library has probably already freed all the memory already which leads to the crash.

johannes-moegerle:
The flathub fix fails in Linux / ubuntu. My guess is, this is due to PyQt5 being on version 5.14 instead of 5.15 as all the other builds.

pairinteraction_gui/worker.py
```diff
@@ -16,6 +16,7 @@
 # along with the pairinteraction GUI. If not, see <http://www.gnu.org/licenses/>.
 from queue import Queue
 
+from PyQt5 import sip
```
hmenke, line 19:
```suggestion
# https://www.riverbankcomputing.com/static/Docs/PyQt5/incompatibilities.html#importing-the-sip-module
try:
    from PyQt5 import sip
except ImportError:
    import sip
```

seweber (approved):

hmenke (changes requested):
Squash the fixup, then merge.

hmenke (approved):

PR #171, opened 2023-10-06
Build with scikit-build-core

hmenke:
- [x] Linux wheels
- [x] macOS wheels
- [ ] Windows wheels
  - [x] `error LNK2019: unresolved external symbol __imp_PyUnicode_AsUTF8 referenced in function "void __cdecl SWIG_Python_AddErrorMsg(char const *)" (?SWIG_Python_AddErrorMsg@@YAXPEBD@Z)`
  - [ ] Install MKL DLLs into the wheel somehow?
- [ ] Currently only `cp38` (which ones to add?)
- [ ] Deploy to PyPI on release?

### DIFF
--- a/.github/workflows/skbuild.yml
+++ b/.github/workflows/skbuild.yml
@@ -1,0 +1,81 @@
+name: skbuild
+
+on:
+  push:
+  pull_request:
+  workflow_call:
+  workflow_dispatch:
+
+# https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+env:
+  CMAKE_BUILD_PARALLEL_LEVEL: 2
+  CTEST_PARALLEL_LEVEL: 2
+  CTEST_OUTPUT_ON_FAILURE: 1
+  VERBOSE: 1
+
+jobs:
+  sdist:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: true
+
+    - name: Build SDist
+      run: pipx run build --sdist
+
+    - name: Check metadata
+      run: pipx run twine check dist/*
+
+    - uses: actions/upload-artifact@v3
+      with:
+        path: dist/*.tar.gz
+
+  wheel:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - os: ubuntu-latest # runs in container anyway
+          CXXFLAGS: "-Wall -Wextra -Wpedantic -march=broadwell"
+          CMAKE_ARGS: "-DBUILD_TESTING=Off"
+        - os: macos-12
+          CXXFLAGS: "-Wall -Wextra -Wpedantic"
+          CMAKE_ARGS: "-DBUILD_TESTING=Off"
+        - os: windows-2019
+          CXXFLAGS: "/Ox /arch:AVX2"
+          CMAKE_ARGS: "-DBUILD_TESTING=Off -DMKL_THREADING=intel_thread"
+    name: Wheel on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: true
+
+    - name: Install Intel oneAPI Math Kernel Library (oneMKL)
+      if: runner.os == 'Windows'
+      uses: ./.github/actions/install-intel-mkl
+
+    - name: Install Windows dependencies
+      if: runner.os == 'Windows'
+      uses: ./.github/actions/setup-vcpkg
+
+    - name: Install macOS dependencies
+      if: runner.os == 'macOS'
+      run: brew install boost cereal doctest eigen fmt gsl lapack
+
+    - uses: pypa/cibuildwheel@v2.16
+      env:
+        CMAKE_ARGS: ${{ matrix.CMAKE_ARGS }}
+        CXXFLAGS: ${{ matrix.CXXFLAGS }}
+        CIBW_MANYLINUX_X86_64_IMAGE: ghcr.io/pairinteraction/pairinteraction-manylinux:docker
+        CIBW_ENVIRONMENT_LINUX: CMAKE_ARGS='${{ matrix.CMAKE_ARGS }}' CXXFLAGS='${{ matrix.CXXFLAGS }}'
+        CIBW_ENVIRONMENT_PASS_LINUX: CMAKE_ARGS CXXFLAGS VERBOSE
+        CIBW_TEST_COMMAND_WINDOWS: "" # wheels are still broken :(
+
+    - name: Upload wheels
+      uses: actions/upload-artifact@v3
+      with:
+        path: wheelhouse/*.whl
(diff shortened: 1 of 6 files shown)


hmenke:
This almost works now. The only remaining this is to find out how to tell CMake to set the RPATH for `_binding.so` to `${PY_MOD_DIR}`. Then cibuildwheel will also be able to properly repair the wheel.

hmenke:
@seweber I think it's just impossible to have dynamically linked libraries inside a wheel. This tiny test project is also not “`auditwheel repair`able” https://github.com/hmenke/test-skbuild-cibuildwheel

We should have a discussion about whether we want to drop `libpairinteraction.so` altogether and only build the `_binding.so`. That would imply removing the `-DWITH_PYTHON` CMake option, because it would make the Python interface mandatory. Another benefit would be that we can remove the `pairinteraction-backend-deprecated`, because it could be replaced with a tiny Python script that only parses arguments and calls the `compute` function.

hmenke:
@seweber I think it's just impossible to have dynamically linked libraries inside a wheel. This tiny test project is also not “`auditwheel repair`able” https://github.com/hmenke/test-skbuild-cibuildwheel

We should have a discussion about whether we want to drop `libpairinteraction.so` altogether and only build the `_binding.so`. That would imply removing the `-DWITH_PYTHON` CMake option, because it would make the Python interface mandatory. Another benefit would be that we can remove the `pairinteraction-backend-deprecated`, because it could be replaced with a tiny Python script that only parses arguments and calls the `compute` function.

seweber:
Would the `_binding.so` also contain the definitions of `libparinteraction.so` so that it could, in principle, be used also by cpp projects? If yes, only building `_binding.so` sounds good to me.

pairinteraction/CMakeLists.txt
```diff
@@ -319,13 +321,15 @@ if(WITH_PYTHON)
     SWIG_COMPILE_OPTIONS "-Wall;-py3"
     SWIG_USE_TARGET_INCLUDE_DIRECTORIES ON
     SWIG_GENERATED_INCLUDE_DIRECTORIES "${Python3_INCLUDE_DIRS};${Python3_NumPy_INCLUDE_DIRS}"
-    SWIG_GENERATED_COMPILE_OPTIONS $<$<CXX_COMPILER_ID:MSVC>:/bigobj>)
+    SWIG_GENERATED_COMPILE_OPTIONS $<$<CXX_COMPILER_ID:MSVC>:/bigobj>
+    INSTALL_RPATH $<$<BOOL:${SKBUILD}>:$<$<PLATFORM_ID:Linux>:$ORIGIN>$<$<PLATFORM_ID:Darwin>:@loader_path>>)
```
hmenke, line 325: This is probably wrong.

pairinteraction/CMakeLists.txt
```diff
@@ -360,14 +364,23 @@ if( WITH_DMG )
...
+      set(PY_MOD_DIR "lib/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/site-packages/pairinteraction")
+    endif()
+
     install(TARGETS binding LIBRARY DESTINATION ${PY_MOD_DIR})
-    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/binding.py DESTINATION ${PY_MOD_DIR})
+    install(TARGETS pairinteraction LIBRARY DESTINATION ${PY_MOD_DIR})
+    install(TARGETS pairinteraction-backend-deprecated RUNTIME DESTINATION ${PY_MOD_DIR})
+    install(FILES ${SWIG_MODULE_binding_SUPPORT_FILES} DESTINATION ${PY_MOD_DIR})
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/pireal.py DESTINATION ${PY_MOD_DIR})
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/picomplex.py DESTINATION ${PY_MOD_DIR})
```
hmenke, line 380: Figure out a better way to install stuff and also which are the correct directories inside the wheel.

pairinteraction/databases/CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ if( WITH_DMG )
 
   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/quantum_defects.db DESTINATION pairinteraction/databases)
 
+elseif( SKBUILD )
+
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/quantum_defects.db DESTINATION .)
```
hmenke, line 36: Probably wrong.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+requires = [
+    "numpy",
+    "pint",
+    "pyqt5",
+    "pyqtgraph",
+    "scipy",
+]
+
+[tool.scikit-build]
+wheel.packages = []
```
hmenke, line 33: Is this even necessary?

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+cmake.verbose = true
+
+[tool.cibuildwheel]
+build = "cp38-manylinux_x86_64"
```
hmenke, line 37: Build for at least a few more Python versions.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+build-verbosity = 1
+manylinux-x86_64-image = "ghcr.io/pairinteraction/pairinteraction-manylinux:docker"
```
hmenke, line 39: Maybe figure out how to vendor most libraries, then we can use any Docker image (use [Conan](https://conan.io/)?).

hmenke (commented):
Looks like it now builds (at least on Linux). Paths inside the wheel are still all fucked up though.

pyproject.toml
hmenke: Also Windows and macOS wheels. Needs restructuring of all the branches for installation.

hmenke:
The Windows build is again a massive time sink. I'm giving up for now.

.github/workflows/skbuild.yml
```diff
@@ -0,0 +1,77 @@
...
+      if: runner.os == 'macOS'
+      run: brew install boost cereal doctest eigen fmt gsl lapack
+
+    - uses: pypa/cibuildwheel@v2.16
+      env:
+        CXXFLAGS: ${{ matrix.CXXFLAGS }}
+        CIBW_MANYLINUX_X86_64_IMAGE: ghcr.io/pairinteraction/pairinteraction-manylinux:docker
+        CIBW_ENVIRONMENT_LINUX: CXXFLAGS='${{ matrix.CXXFLAGS }}'
+        CIBW_ENVIRONMENT_PASS_LINUX: CXXFLAGS VERBOSE
+        CIBW_TEST_COMMAND_WINDOWS: "" # wheels are still broken :(
```
hmenke, line 72: Don't be fooled by any passing CI. The wheel is still broken, just the test is disabled.

johannes-moegerle:
> ```shell
> pip install --verbose git+https://github.com/hmenke/pairinteraction.git@scikit-build
> ```

I was trying this in a python3.11 virtual environment (to not screw up my current python version I was only installing python3.11-venv) via
```shell
sudo apt install python3.11-venv
python3.11 -m venv myenv
source ./myenv/bin/activate
```

Now running the pip install resulted in an error
```
Could NOT find Python3 (missing: Development.Module NumPy) (found version
    "3.11.6")

        Reason given by package:
            Development: Cannot find the directory "/usr/include/python3.11"
```
So I guess cmake still was trying to use some global python3.11.
I also found and tried this: https://discourse.cmake.org/t/possible-to-create-a-python-virtual-env-from-cmake-and-then-find-it-with-findpython3/1132 but did not manage to get it to work.

Not sure if this is a problem we want to care about, or someone knows a simple fix? Just wanted to let you know.

Using a conda python3.11 environment seems to work without any problems.

PR #204, opened 2025-03-13
Improve contributor guide, cleanup logging and dependencies

seweber:


docs/contribute/getting_started.rst
```diff
@@ -56,7 +58,15 @@ functioning on your computer. In the following we provide a brief summary of the
     In addition, you need to install the dependencies of the C++ backend: If you are
     using GNU/Linux or OS X, dependencies can be found in the Dockerfiles that are located in the :github:`docker
     branch <tree/docker/docker>`. If you are using Windows, you can use vcpkg_ with :github:`our configuration file
-    <tree/master/vcpkg.json>` to install the dependencies.
+    <tree/master/vcpkg.json>` to install the dependencies. You can do so by running the following commands in the root
```
EduardJBraun, line 61: This has the possible drawback that the user must install vcpkg in his pairinteraction directory. I would, especially on Windows, rather recommend to add vcpkg to the enivronment variables, and then only use vcpkg install --triplet x64-windows

docs/contribute/getting_started.rst
```diff
@@ -84,8 +94,8 @@ repository with the following command:
 
     pre-commit install
 
-This automatically formats your code and conducts style checks before each commit. For manual checks at any time,
-execute:
+This automatically formats your code and conducts style checks before each commit. On Windows, you might want to set `$env:SKIP = "clang-format"`
```
EduardJBraun, line 97: On my windows system, I actually never had a problem with clang-format. Is this maybe already automatically set up

EduardJBraun (approved):

src/pairinteraction/logging.py
```diff
@@ -67,12 +67,7 @@ def decorate_module_with_flush_logs(module: object) -> None:
 
 def configure_logging(
     level_str: str = "WARNING",
-    fmt: str = (
-        f"[{Fore.MAGENTA}%(asctime)s.%(msecs)03d{Style.RESET_ALL}] "
-        f"[%(levelname)s] "
-        f"[{Fore.CYAN}%(filename)s:%(lineno)d{Style.RESET_ALL}] "
-        f"%(message)s"
-    ),
+    fmt: str = ("[%(asctime)s.%(msecs)03d] [%(levelname)s] [%(filename)s:%(lineno)d] %(message)s"),
```
johannes-moegerle, line 70: wasnt the main advantage to have colorful logs, how does this now work with the new fmt?

johannes-moegerle (approved):

src/pairinteraction/logging.py
johannes-moegerle: No I agree, having a consistent color is better.
I just wasn't sure which color is set where.

PR #205, opened 2025-03-13
More convenience and numerical stability of the perturbative functions

EduardJBraun:
For calcualtion of the C6, feature was added that only certain order corrections can be calculated and returned. Enhances stability especially for large distances. Also automatic system creation was added.

### DIFF
--- a/src/pairinteraction/perturbative/perturbative.py
+++ b/src/pairinteraction/perturbative/perturbative.py
@@ -664,145 +664,71 @@ def _create_system(
     if electric_field.magnitude[1] != 0 or magnetic_field.magnitude[1] != 0:
         pi = pi_complex
 
-    n_1 = []
-    l_1 = []
-    j_1 = []
-    m_1 = []
-    energ_au = []
-    n_2 = []
-    l_2 = []
-    j_2 = []
-    m_2 = []
-    for ket1, ket2 in ket_tuple_list:
-        n_1.append(ket1.n)
-        l_1.append(ket1.l)
-        j_1.append(ket1.j)
-        m_1.append(ket1.m)
-        n_2.append(ket2.n)
-        l_2.append(ket2.l)
-        j_2.append(ket2.j)
-        m_2.append(ket2.m)
-        energ_au.append(ket1.get_energy().magnitude + ket2.get_energy().magnitude)
-    species = [ket_tuple_list[0][0].species, ket_tuple_list[0][1].species]
-    n_max = [np.max(n_1), np.max(n_2)]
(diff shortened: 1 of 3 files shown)


seweber:
Capital letters in variable names are rather uncommon (see also the [report](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=205&id=pairinteraction_pairinteraction&open=AZWQAqa8nKpiwbOGcitM&tab=why) by sonarqube). Thus, instead of `deltaE`, `H` better use `delta_energy`, `hamiltonian`.

src/pairinteraction/perturbative/perturbative.py
```diff
@@ -378,3 +622,169 @@ def _check_for_resonances(
...
+    )
+    dipole_min_2 = pi.KetAtom(
+        species[1], n=n_min_2, l=n_min_2 - 1, j=n_min_2 - 1 + spin[1], m=n_min_2 - 1 + spin[1]
+    ).get_matrix_element(
+        pi.KetAtom(species[1], n=n_min_2 + 1, l=n_min_2, j=n_min_2 + spin[1], m=n_min_2 + spin[1]),
+        operator="ELECTRIC_DIPOLE",
+        q=1,
+    )
+    shift_min = dipole_min_1 * dipole_min_2 * ureg.coulomb_constant / distance**3
+    delta_energy_min = np.abs(shift_min) / np.sqrt(population_admixture) * perturbative_order
```
johannes-moegerle, line 777: the calculation of delta_energy_min and delta_energy_max is rather complicated, and i dont really understand what is going on there.
I shortly tested this with this ket_tuple_list
```
ket = pi_real.KetAtom("Rb", n=30, l=0, m=0.5)
ket_tuple_list = [(ket, ket)]
```
and got delta energies of 0.08GHz. I dont think that this is what we want.
On the other hand even for n=100, the delta energies are only ~10GHz.
So I wonder if there is not a simpler solution like:
1.  just use some constant delta_energy (e.g. 20GHz, this will probably still result in more or less instant perturbative calculations).
2. Or maybe a bit better, have some delta_energy = constant * n^X, where we again think about what a sensible exponent X could be (do we also expect a l dependence?)
3. Maybe one could also just always try to create a basis with roughly a fixed number of kets, but since this is really rather unphysical I dont necessarily prefer this. However, one could at the end do a sanity check, whether the basis is neither too large, nor too small.

johannes-moegerle (changes requested):

PR #206, opened 2025-03-13
Fix linting

seweber:


tests/test_ket.py
```diff
@@ -6,5 +8,5 @@ def test_ket() -> None:
     assert ket.species == "Rb"
     assert ket.n == 60
     assert ket.l == 0
-    assert ket.j == 0.5
-    assert ket.m == 0.5
+    assert pytest.approx(ket.j) == 0.5  # NOSONAR
```
johannes-moegerle, line 11: why doesnt sonar like this?

tests/test_perturbative.py
```diff
@@ -52,7 +52,7 @@ def _create_system_pair_sample():
             l=(0, 1),
             m=(-1.5, 1.5),
         )
-        for i in [1, 2]
+        for _ in [1, 2]
```
johannes-moegerle, line 55: if it is twice the same basis and twice the same system we should only create on, diagonalize one, and then call BasisPair([system, system])

tests/test_perturbative.py
```diff
@@ -178,7 +178,7 @@ def test_resonance_detection(caplog):
     with pytest.warns(RuntimeWarning, match="divide by zero encountered in divide"):
         with pytest.raises(ValueError) as excinfo:
             with caplog.at_level(logging.CRITICAL):
-                H_eff, _ = perturbative.get_effective_hamiltonian_from_system(
+                _, _ = perturbative.get_effective_hamiltonian_from_system(
```
johannes-moegerle, line 181: just remove the _, _ = completely and only call the function

tests/test_ket.py
johannes-moegerle: would sonar be more happy with `np.isclose` or `np.testing.assert_allclose`?
also for below in the other tests

src/cpp/src/basis/Basis.cpp
```diff
@@ -388,7 +390,8 @@ Basis<Derived>::get_rotator(real_t alpha, real_t beta, real_t gamma) const {
     for (size_t idx_initial = 0; idx_initial < kets.size(); ++idx_initial) {
         real_t f = kets[idx_initial]->get_quantum_number_f();
         real_t m_initial = kets[idx_initial]->get_quantum_number_m();
-        for (real_t m_final = -f; m_final <= f; ++m_final) {
+        for (int delta = 0; delta <= 2 * f + precision; ++delta) {
```
johannes-moegerle, line 393: why is this better then the previous expression?

tests/test_ket.py
seweber: It's an issue with the linter of sonar: The linter thinks that we are comparing two floats with "==" and is not realizing that we are comparing two floats approximately.  In contrast to the opinion of the linter, our code is perfectly fine. Thus, I have switched sonar off for the code lines (with sonar's "Automatic Analysis Method" that we are using, I have not found a way to switch off the error-prone check globally).

johannes-moegerle: i see, always putting nosonar is still kind of annoying, which would be an argument to use np.isclose, but I dont really mind

src/cpp/src/basis/Basis.cpp
johannes-moegerle: ok, then leave it as is now

johannes-moegerle (approved):

src/cpp/src/basis/Basis.cpp
seweber: I thought again about it. If f and m are not half integers, the algorithms are not working. Thus, having asserts is anyway a good idea. In this case, directly looping over m_final might be the cleaner solution. I adapted the code accordingly.

src/cpp/src/basis/Basis.cpp
```diff
@@ -388,7 +389,12 @@ Basis<Derived>::get_rotator(real_t alpha, real_t beta, real_t gamma) const {
     for (size_t idx_initial = 0; idx_initial < kets.size(); ++idx_initial) {
         real_t f = kets[idx_initial]->get_quantum_number_f();
         real_t m_initial = kets[idx_initial]->get_quantum_number_m();
-        for (real_t m_final = -f; m_final <= f; ++m_final) {
+
+        assert(2 * f == std::floor(2 * f) && f >= 0);
+        assert(2 * m_initial == std::floor(2 * m_initial) && m_initial >= -f && m_initial <= f);
+
+        for (real_t m_final = -f; m_final <= f; // NOSONAR m_final is precisely representable
```
johannes-moegerle, line 396: for now it is fine, but I would like if we do not have to write too often NOSONAR ...

johannes-moegerle (approved):

PR #207, opened 2025-03-14
Fix the check of the precision of eigenvectors and eigenvalues

seweber:


src/cpp/src/interfaces/DiagonalizerInterface.cpp
```diff
@@ -34,38 +34,44 @@ DiagonalizerInterface<Scalar>::subtract_mean(const Eigen::MatrixX<Scalar> &matri
...
-
-    if (error_from_float_type * 1e1 > error_from_eigenvectors) { // 1e1 is a safety factor
-        SPDLOG_WARN("Because the floating point precision is too low, the "
-                    "eigenvalues cannot be calculated accurately. The floating point error ({} "
-                    "Hartree) is similar or larger than error estimated from the specified "
-                    "tolerance ({} Hartree).",
-                    error_from_float_type, error_from_eigenvectors);
-    }
+    // eigenvector matrix that are smaller than the tolerance are set to zero:
+    // |Tr(D_exact) - Tr(D)|/dim(D) <= 2*atol*||shifted_matrix||
```
johannes-moegerle, line 63: isnt the trace kind of a big lie? should one take somthing like the maximum difference of the entries?

johannes-moegerle, line 63: isnt the trace kind of a lie? should one take something like the maximum difference of the entries?

johannes-moegerle: also atol * ... seems wrong when thinking of atol as absolute tolerance

johannes-moegerle (commented):
assuming all the arguments are correct (I think we have to discuss this shortly) the final result is we dont need any check apart from checking that:
`error_from_float_type <= atol` ?
In this case I wonder why are we doing this comparison, both variables are basically a user input, so we have two inputs from the user, just to check them and throw an error if they are not consistent.
Is atol actually used for something else as well? Or could we just induce whether to use float32 or float64 from the given atol?

src/cpp/src/interfaces/DiagonalizerInterface.cpp
seweber: I expect that it is tough to get a bound for the maximum difference of the entries. The trace error has the advantage that, because of the nice properties of the trace, bounds can be achieved rather easily. To my understanding, an expression similar to the trace error is also used by FEAST to determine whether it is converged, see https://arxiv.org/pdf/2002.04807, table 3 on page 17.

seweber: > also atol * ... seems wrong when thinking of atol as absolute tolerance

It depends. With respect to the eigenvectors, atol behaves as an absolute tolerance. For the eigenvalues it acts as a relative tolerance (so, maybe, we should just call it "tol"?).

> Is atol actually used for something else as well? Or could we just induce whether to use float32 or float64 from the given atol?

For FEAST and EVR, atol is used to specify the precision.

seweber: It looks like I was too pessimistic concerning the ability to get better error estimates than from the trace norm. It seems to work out nicely to bound the errors of the eigenvalues |lambda - lambda_exact| by a user-specified "rtol" times the norm of the Hamiltonian matrix. Now, everything is specified in terms of rtol. Also, the decision on which entries in the eigenvector matrix are discarded is taken based on the value of rtol.

src/cpp/src/interfaces/DiagonalizerInterface.cpp
```diff
@@ -18,54 +18,60 @@ template <typename Scalar>
...
-    if (error_from_float_type * 1e1 > atol) { // 1e1 is a safety factor
-        SPDLOG_WARN("Because the floating point precision is too low, the "
-                    "eigenvectors cannot be calculated accurately. The floating point error ({}) "
-                    "is similar or larger than the specified tolerance ({}).",
-                    error_from_float_type, atol);
+    if (floating_point_error > rtol) {
+        SPDLOG_WARN(
+            "Because the floating point precision is too low, the eigenvalues cannot be calculated "
+            "accurately. The estimated floating point error ({} * ||H||) "
+            "is larger than the specified tolerance ({} * ||H||).",
```
johannes-moegerle, line 40: add a comment, what the user should do: either consider increasing rtol or changing float type

johannes-moegerle (approved):

PR #208, opened 2025-03-14
Implement interaction with the field on an ion

seweber:


src/cpp/include/pairinteraction/utils/spherical.hpp
```diff
@@ -23,21 +23,67 @@ convert_to_spherical_basis(const std::array<typename traits::NumTraits<Scalar>::
...
     }
 
+    std::array<Scalar, 3> field_spherical{};
+    Eigen::Map<Eigen::Vector3<Scalar>>(field_spherical.data(), field_spherical.size()) =
+        converter.template cast<Scalar>() *
+        Eigen::Map<const Eigen::Vector3<real_t>>(field.data(), field.size());
     return field_spherical;
 }
+
+template <typename Scalar, int Order>
```
johannes-moegerle, line 44: is this really the way you do this in C++ do template the order parameter? seems kind of awkward

src/pairinteraction/_wrapped/system/SystemAtom.py
```diff
@@ -84,6 +84,28 @@ def enable_diamagnetism(self: "Self", enable: bool = True) -> "Self":
...
+        return self.set_distance_vector_to_ion([0, 0, distance], unit)
+
+    def set_distance_vector_to_ion(
+        self: "Self",
+        distance: Union["PlainQuantity[NDArray[Any]]", Collection[Union[float, "PlainQuantity[float]"]]],
+        unit: Optional[str] = None,
+    ) -> "Self":
+        distance_au = [QuantityScalar.from_pint_or_unit(v, unit, "DISTANCE").to_base_unit() for v in distance]
+        self._cpp.set_distance_vector_to_ion(distance_au)
+        self._distance_vector_au = distance_au
```
johannes-moegerle, line 99: as long as we dont provide a get_ion_distance i dont think we need `self._distance_vector_au`, or implement an get_ion_distance method ;)

src/cpp/include/pairinteraction/utils/spherical.hpp
seweber: Yes, it must be a template parameter because the order is used as a template parameter to set the dimension of std::array at compile time. Alternatively, one could use a std:.vector whose size can be adapted at runtime, but I think it's better to do it at compile time (better type safety and performance).

johannes-moegerle: I understand this, I was just wondering, since the code for both cases is rather different, whether this function makes sense as one function. Also can this be reused for something else then ion interactions?
Otherwise, maybe just put this code to constructing the ion Hamiltonian?

seweber: Ah, yes, I also was not too happy with this code and now refactored it significantly. Now, there is one function, "convert_to_spherical_basis", that is used for both multipole interaction and interaction with fields. I am just not sure what the best conventions are, i.e., where to put which prefactors and where to take the complex conjugate of spherical harmonics. There are multiple mathematically equivalent ways how it can be done. We can discuss it next week.

seweber: In the long run, we could think about extending the "convert_to_spherical_basis" function to also work for higher orders than 1 and 2.

seweber: Now, I am directly using the "get_transformator" method. I think this might be the cleanest solution.

src/cpp/src/system/SystemAtom.cpp
```diff
@@ -52,57 +113,84 @@ void SystemAtom<Scalar>::construct_hamiltonian() const {
     bool sort_by_quantum_number_m = true;
     bool sort_by_parity = true;
 
-    // Add external fields (see https://arxiv.org/abs/1612.08053 for details)
```
johannes-moegerle, line 55: is there a reason why you removed the reference to the old pairinteraction paper, probably it is not needed, but i feel like it also does not hurt

src/cpp/src/system/SystemAtom.cpp
```diff
@@ -52,57 +113,84 @@ void SystemAtom<Scalar>::construct_hamiltonian() const {
...
-        sort_by_parity = false;
+
+    // Stark effect: - \vec{d} \vec{E} = - d_{1,0} E_{0} + d_{1,1} E_{-} + d_{1,-1} E_{+}
+    // = - d_{1,0} E_{0} - d_{1,1} E_{+}^* - d_{1,-1} E_{-}^*
+    // with the electric dipole operator: d_{1,q} = - e r sqrt{4 pi / 3} Y_{1,q}(\theta, \phi)
+    // where electric_field_spherical=[E_{-}^*, E_{0}, E_{+}^*]
+
+    for (int q = -1; q <= 1; ++q) {
+        if (std::abs(electric_field_spherical.at(q + 1)) > numerical_precision) {
+            *this->hamiltonian -= electric_field_spherical.at(q + 1) *
```
johannes-moegerle, line 172: Why is there this asymmetry of the datatype of `electric_field_spherical` and e.g. `ion_first_order`, seems like they could be the same?

src/cpp/src/system/SystemAtom.cpp
```diff
@@ -122,30 +212,30 @@ void SystemAtom<Scalar>::construct_hamiltonian() const {
...
             this->hamiltonian_is_diagonal = false;
             sort_by_quantum_number_f = false;
         }
         if (std::abs(magnetic_field_spherical[1]) > numerical_precision &&
-            std::abs(magnetic_field_spherical[0]) > numerical_precision) {
-            *this->hamiltonian += static_cast<real_t>(std::sqrt(3.0) / 12.) *
-                magnetic_field_spherical[1] * magnetic_field_spherical[0] *
+            std::abs(magnetic_field_spherical[2]) > numerical_precision) {
+            *this->hamiltonian -= static_cast<real_t>(std::sqrt(3.0) / 12.) *
+                magnetic_field_spherical[1] * magnetic_field_spherical[2] *
```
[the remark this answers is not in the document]
johannes-moegerle: got it, nvm

src/cpp/src/system/SystemAtom.cpp
```diff
@@ -154,14 +244,6 @@ void SystemAtom<Scalar>::construct_hamiltonian() const {
...
             *this->hamiltonian -= static_cast<real_t>(std::sqrt(1.5) / 12.) *
                 static_cast<Scalar>(std::pow(magnetic_field_spherical[0], 2)) *
-                OperatorAtom<Scalar>(basis, OperatorType::ELECTRIC_QUADRUPOLE, 2);
-            this->hamiltonian_is_diagonal = false;
-            sort_by_quantum_number_f = false;
-            sort_by_quantum_number_m = false;
-        }
-        if (std::abs(magnetic_field_spherical[2]) > numerical_precision) {
-            *this->hamiltonian -= static_cast<real_t>(std::sqrt(1.5) / 12.) *
-                static_cast<Scalar>(std::pow(magnetic_field_spherical[2], 2)) *
```
johannes-moegerle, line 164: why is this term removed? (if it is the same as the term above, should the prefactor has changed?)

seweber: Ah, thanks. Yes, this term should not have been removed.

tests/test_ion.py
```diff
@@ -0,0 +1,39 @@
...
+        for d in distances
+    ]
+    systems_z = [
+        pi.SystemAtom(basis).set_ion_charge(1, unit="e").set_distance_vector_to_ion([0, 0, d], unit="um")
+        for d in distances
+    ]
+
+    # Diagonalize the systems in parallel
+    pi.diagonalize(systems_x, diagonalizer="eigen", sort_by_energy=True)
+    pi.diagonalize(systems_y, diagonalizer="eigen", sort_by_energy=True)
```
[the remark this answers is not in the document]
seweber: Good idea. I now did some additional improvements to the test case and also to the test of the diamagnetic interaction so that more errors can be caught in the future.

seweber:
I had to rebase on master in order to get the CI running. The commits [0f5576c](https://github.com/pairinteraction/pairinteraction/pull/208/commits/0f5576cf60bc8f1fc652cbb9673f8a6c6ebaac65) and [1a56af7](https://github.com/pairinteraction/pairinteraction/pull/208/commits/1a56af766f7184358ff793693f319b206086823e) contain the changes made after the review.

tests/test_ion.py
```diff
@@ -31,7 +31,7 @@ def test_ion() -> None:
     eigenenergies = system_x.get_eigenenergies(unit="GHz")
     overlaps = system_x.basis.get_overlaps(ket)
     idx = np.argmax(overlaps)
-    assert pytest.approx(overlaps[idx], rel=1e-12) == 0.8754157131466024  # NOSONAR
+    assert pytest.approx(overlaps[idx], rel=0.15) == 0.8754157131466024  # NOSONAR
```
johannes-moegerle, line 34: why do you need such a big error tolerance? is the value wrong or is this again an artifact of the smaller basis for the tests? maybe add a comment to clarify this

johannes-moegerle (approved):

PR #212, opened 2025-03-19
Precommit/check rst sphinx

johannes-moegerle:
Add a pre-commit to check / format rst files for sphinx

I tried the following repos:
- [sphinx-lint](https://github.com/sphinx-contrib/sphinx-lint) 
  - actually no idea what they do check
- [rstcheck](https://github.com/rstcheck/rstcheck)
  - does not seem to play to nicely with sphinx and reports to many false positives, which is annoying
- [doc8](https://github.com/PyCQA/doc8)
  - the only linter, which supports checking the line length, seems to work ok
- [docstrfmt](https://github.com/LilSpazJoekp/docstrfmt)
  - the only useful formater i found, can be a bit slow, unfortunetaly does not support rubric sections

Thus, I would propose to use doc8 and docstrfmt

## TODO
Not sure yet whether to only add docstrfmt as manual precommit hook are always running it?


johannes-moegerle:
add autofix commit to git blame ignore

seweber (approved):

PR #214, opened 2025-03-20
Various fixes

johannes-moegerle:
Addressing the following issues:
- https://github.com/atom-pairinteraction/pairinteraction/issues/183
- https://github.com/atom-pairinteraction/pairinteraction/issues/170
- https://github.com/atom-pairinteraction/pairinteraction/issues/160
- a bug in the gui, if the databases were not yet downloaded
- add efield sensitivty notebook again


seweber:
I agree that set_distance is not needed in the cpp code. You can remove/replace it with set_distance_vector everywhere it occurs in src/cpp.

.github/workflows/python-wheel.yml
```diff
@@ -110,13 +110,18 @@ jobs:
     - name: Install the wheel
       run: |
         python -m pip install --upgrade pip
-        python -m pip install --find-links=dist pairinteraction
+        python -m pip install --find-links=dist pairinteraction[tests]
 
     - name: Execute a test that downloads the database and runs checks
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: python -c "import sys; from pairinteraction import run_module_tests; sys.exit(run_module_tests(download_missing=True))"
```
seweber, line 118: Should we run "pairinteraction --log-level INFO test" instead of "import sys; from pairinteraction import run_module_tests; sys.exit(run_module_tests(download_missing=True))"?

docs/tutorials/examples_python/perturbative_h_eff.ipynb
```diff
@@ -13,7 +13,7 @@
    "source": [
     "This notebook demonstrates how to calculate an effective Hamiltonian with the `pairinteraction.pertrubative` module.\n",
```
seweber, line 14: pertrubative -> perturbative

src/pairinteraction/cli.py
```diff
@@ -46,6 +46,10 @@ def main() -> int:
     purge_parser = subparsers.add_parser("purge", help="delete all cached data")
     purge_parser.set_defaults(func=lambda args: purge_cache())
 
+    if len(sys.argv) == 1:
```
seweber, line 49: It looks like there should be a proper way to achieve it with argparse. Maybe, there is something as a "default" keyword.

src/pairinteraction_gui/__main__.py
```diff
@@ -1,4 +1,30 @@
...
+
+
+def main() -> int:
+    """Run the PairInteraction GUI application.
+
+    Returns:
+        int: Application exit code
+
+    """
+    # in case we run the app from the terminal, add some logging
```
seweber, line 15: I would forbid it to start the GUI differently than by running "pairinteraction gui". Otherwise, we have two places where we have to maintain setting up the logging etc.

src/pairinteraction_gui/app.py
```diff
@@ -1,64 +1,56 @@
...
+    @classmethod
+    def instance(cls) -> "Application":  # type: ignore  # overwrite type hints
+        """Return the current instance of the application."""
+        return super().instance()  # type: ignore [return-value]
 
-    def signal_handler(signal: int, frame: Optional[FrameType]) -> None:
-        os.write(pipe_w, b"x")  # Write a single byte to the pipe
+    def allow_ctrl_c(self) -> None:
+        # # Make program killable via Ctrl+C if it is started in a terminal
+        # signal.signal(signal.SIGINT, signal.SIG_DFL)  # this kills the program
```
seweber, line 33: Remove "# #" and comment out code.

src/pairinteraction_gui/main_window.py
```diff
@@ -225,3 +234,19 @@ def init_keyboard_shortcuts(self) -> None:
         # Add Ctrl+W shortcut to close the window
         close_shortcut = QShortcut(QKeySequence("Ctrl+W"), self)
         close_shortcut.activated.connect(self.close)
+
+    def ask_download_database(self, species: str) -> bool:
+        msg_box = QMessageBox()
+        msg_box.setWindowTitle("Download missing databases?")
+        msg_box.setText(f"Databases for {species} not found.")
```
seweber, line 241: Databases -> Database tables

src/pairinteraction_gui/__main__.py
johannes-moegerle: i agree, that we should not handle logging here and removed it, however i do not strictly forbid to start the gui via `python -m pairinteraction_gui` since i dont think it hurts for now (and we anyway dont document this).
But this makes the package `pairinteraction_gui` more standalone and less intertwined with the pairinteraction package

docs/tutorials/examples_python/perturbative_h_eff.ipynb
johannes-moegerle: done

.github/workflows/python-wheel.yml
johannes-moegerle: agree, on that note currently the cli command `pairinteraction test` is the only command, where we implicitly download stuff, is this ok or should we think about a `--download-missing` flag for the cli as well?

src/pairinteraction_gui/config/system.py
```diff
@@ -206,13 +206,13 @@ def setupWidget(self) -> None:
         self.items = []
 
         min_spinbox = DoubleSpinBox(
-            self, vmin=0, vmax=999, vdefault=3, tooltip=f"Minimum distance in {parse_html('mu')}"
+            self, vmin=0, vmax=999, vdefault=3, tooltip=f"Minimum distance in {parse_html('mu')}m"
```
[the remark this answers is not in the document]
johannes-moegerle: micrometer, clearified this

src/pairinteraction/cli.py
johannes-moegerle: changed it to
`args = parser.parse_args(args=None if sys.argv[1:] else ["--help"])`

johannes-moegerle:
> I agree that set_distance is not needed in the cpp code. You can remove/replace it with set_distance_vector everywhere it occurs in src/cpp.

done

src/pairinteraction_gui/main_window.py
```diff
@@ -225,3 +234,19 @@ def init_keyboard_shortcuts(self) -> None:
         # Add Ctrl+W shortcut to close the window
         close_shortcut = QShortcut(QKeySequence("Ctrl+W"), self)
         close_shortcut.activated.connect(self.close)
+
+    def ask_download_database(self, species: str) -> bool:
+        msg_box = QMessageBox()
+        msg_box.setWindowTitle("Download missing databases?")
```
seweber, line 240: databases -> database tables (I tried to use the convention that with "database", the full storage of matrix elements and states is meant. "database tables" are the individual parquet files)

seweber (changes requested):

src/pairinteraction/cli.py
seweber: Hmm, sys.argv[1:] looks like a hack to me as we are using argparse, and there should be some built-in solution.

src/pairinteraction_gui/main_window.py
johannes-moegerle: ah i see, changed it now also in the title

src/pairinteraction/cli.py
johannes-moegerle: i think the build in solution would be to make command optional and  set its default to help and add an additional help command, but i dont think i like this more, another way is subclassing `argparse.ArgumentParser` and override the error method to always also print the help when some argparse error happens.
However, both solutions are rather complex for simply printing a help message, if no argument was passed

seweber: You could try "parser.set_defaults(func=lambda args: parser.print_help())" or "if args.command is None: ...".

.github/workflows/python-wheel.yml
seweber: Good question, for me, both would be fine.

src/pairinteraction/cli.py
johannes-moegerle: but then you have to make the command optional, not sure if this is better?

johannes-moegerle: I now made required False, and check `if not args.command: ...`

seweber (approved):

PR #225, opened 2025-04-02
Lint the source code of the tools

seweber:
TODOs
<del>- Check that the dbmanager is still working</del> done
<del>- Let the dbmanager download only species relevant for the unit tests</del> done


CMakeLists.txt
```diff
@@ -153,6 +152,21 @@ if(WIN32)
   set_target_properties(duckdb PROPERTIES IMPORTED_IMPLIB ${duckdb_SOURCE_DIR}/duckdb.lib)
 endif()
 
+# Doctest
+# We use the dev version because it works with recent cmake versions that forbid a version < 3.5 for cmake_minimum_required
+FetchContent_Declare(
+  doctest
+  GIT_REPOSITORY https://github.com/doctest/doctest
+  GIT_TAG 3a01ec37828affe4c9650004edb5b304fb9d5b75 # v0.20.0
```
hmenke, line 160: First of all, this commit does not point to a release tag, and even if it were, it's surely not v0.20.0, because that never existed.
```suggestion
  GIT_TAG 1da23a3e8119ec5cce4f9388e91b065e20bf06f5 # v2.4.12
```

PR #226, opened 2025-04-03
Gui/multiprocess

johannes-moegerle:
merge after ##223

### DIFF
--- a/src/pairinteraction_gui/app.py
+++ b/src/pairinteraction_gui/app.py
@@ -1,13 +1,17 @@
 # SPDX-FileCopyrightText: 2025 Pairinteraction Developers
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
+import logging
 import os
 import signal
+from multiprocessing import Process
 from types import FrameType
-from typing import Optional
+from typing import ClassVar, Optional
 
-from PySide6.QtCore import QCoreApplication, QObject, QSocketNotifier, QTimer, Signal
+from PySide6.QtCore import QObject, QSocketNotifier, QThread, QTimer, Signal
 from PySide6.QtWidgets import QApplication
+
+logger = logging.getLogger(__name__)
 
 
 class MainSignals(QObject):
@@ -25,6 +29,8 @@ class Application(QApplication):
     """Add some global signals to the QApplication."""
 
     signals = MainSignals()
+    all_processes: ClassVar[set[Process]] = set()
+    all_threads: ClassVar[set[QThread]] = set()
 
     @classmethod
     def instance(cls) -> "Application":  # type: ignore  # overwrite type hints
@@ -42,8 +48,8 @@ class Application(QApplication):
 
         def handle_signal() -> None:
             os.read(pipe_r, 1)  # Read the byte from the pipe to clear it
-            print("\nCtrl+C detected. Shutting down gracefully...")
-            QCoreApplication.quit()
+            logger.info("Ctrl+C detected in terminal. Shutting down gracefully...")
+            self.quit()
 
         sn = QSocketNotifier(pipe_r, QSocketNotifier.Type.Read, parent=self)
         sn.activated.connect(handle_signal)
@@ -53,3 +59,23 @@ class Application(QApplication):
         timer = QTimer(self)
         timer.timeout.connect(lambda: None)  # Do nothing, just wake up the event loop
         timer.start(200)
+
+    @staticmethod
+    def quit() -> None:
+        """Quit the application."""
+        logger.debug("Calling Application.quit().")
+
+        for process in Application.all_processes:
+            if process.is_alive():
+                logger.debug("Terminating process %s.", process.pid)
+                process.terminate()
+                process.join(timeout=1)
+
+        for thread in Application.all_threads:
+            if thread.isRunning():
+                logger.debug("Terminating thread %s.", thread)
+                thread.terminate()
+                thread.wait()
+
+        QApplication.quit()
+        logger.debug("Application.quit() done.")
--- a/src/pairinteraction_gui/calculate/calculate_base.py
+++ b/src/pairinteraction_gui/calculate/calculate_base.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2025 Pairinteraction Developers
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
+import logging
 from abc import ABC
 from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar, Union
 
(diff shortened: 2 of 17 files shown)


src/pairinteraction_gui/app.py
```diff
@@ -53,3 +56,27 @@ def handle_signal() -> None:
         timer = QTimer(self)
         timer.timeout.connect(lambda: None)  # Do nothing, just wake up the event loop
         timer.start(200)
+
+    @staticmethod
+    def quit() -> None:
+        """Quit the application."""
+        logger.debug("Calling Application.quit().")
+        from pairinteraction_gui.worker import ALL_PROCESSES, ALL_THREADS
```
seweber, line 64: Could the ALL_PROCESSES and ALL_THREADS be member variables of the Application class (and then dependency injection used for the workers)? I am not sure whether my idea works out, but if possible without much overhead, I would avoid the global variables.

src/pairinteraction_gui/page/base_page.py
seweber, line 1: base_page or page_base (analogously to calculate_base)? In case of the letter, I would also use page_lifetimes etc.

src/pairinteraction_gui/calculate/calculate_base.py
seweber, line 1: I wonder whether a calculate_lifetimes would also belong into the calculate folder. Then, the calculate folder would have the same structure as the page folder.

src/pairinteraction_gui/calculate/calculate_two_atoms.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@dataclass
+class ResultsTwoAtoms(Results):
+    basis_0_label: Optional[str] = None
+
+
+@run_in_other_process
+def calculate_two_atoms(parameters: ParametersTwoAtoms) -> ResultsTwoAtoms:
+    """Calculate the energy plot for one atom.
```
seweber, line 43: for one atom -> for two atoms

src/pairinteraction_gui/main_window.py
```diff
@@ -236,8 +236,18 @@ def init_keyboard_shortcuts(self) -> None:
         """Initialize keyboard shortcuts."""
         # Add Ctrl+W shortcut to close the window
         close_shortcut = QShortcut(QKeySequence("Ctrl+W"), self)
+        close_shortcut.activated.connect(lambda: logger.info("Ctrl+W detected. Shutting down gracefully..."))
         close_shortcut.activated.connect(self.close)
 
+    def close(self) -> bool:
+        """Quit the complete application.
+
+        Call Application.quit() instead of super().close(),
```
seweber, line 245: There is nothing that super().close() would do but Application.quit() does not do?

src/pairinteraction_gui/page/base_page.py
```diff
@@ -87,37 +74,85 @@ def hideEvent(self, event: QHideEvent) -> None:
...
+        super().setupWidget()
+
+        # Control panel below the plot
+        control_layout = QHBoxLayout()
+        calculate_button = QPushButton("Calculate")
+        calculate_button.setObjectName("Calculate")
+        calculate_button.clicked.connect(self.calculate_clicked)
+        control_layout.addWidget(calculate_button)
+
+        # export_button = QPushButton("Export")
```
seweber, line 93: Remove commented-out code or add a TODO / FIXME to explain what should be done with the code in the future.

src/pairinteraction_gui/plotwidget/plotwidget.py
```diff
@@ -72,10 +69,14 @@ class PlotEnergies(PlotWidget):
...
-        max_spinbox = DoubleSpinBox(self, vmin=-999, vmax=999, vdefault=0.5, decimals=2)
         self.energy_range = RangeItem(
-            self, "Calculate the energies from", min_spinbox, max_spinbox, "GHz", checked=False
+            self,
+            "Calculate the energies from",
+            (-999, 999),
+            (-0.5, 0.5),
+            unit="GHz",
+            checked=False,
+            long_label="energy",
```
seweber, line 79: If "energy" is the "long_label", what is a short label?

src/pairinteraction_gui/page/base_page.py
johannes-moegerle: for page and config i had the convention `<type>_page.py` and `<type>_config.py` which then defined the `<Type>Config` and `<Type>Page` classes, where type also include base (so e.g. `base_config.py` defines `BaseConfig` class).
The confusion probably comes with the calculate functions, which i wanted to call `def calculate_<type>`, and thus also named the files `calculate_<type>.py` (where `calculate_base.py` however only defines Parameters and Results, since there is no `calculate_base` function).
I see that this might be a bit confusing, so if you have suggestions how to improve this naming conventions i would be glad to talk about this.

src/pairinteraction_gui/plotwidget/plotwidget.py
johannes-moegerle: haha, yeah the first argument was the short label, i renamed short_label -> label and long_label -> tooltip_label which now describes their purposes more accurately

src/pairinteraction_gui/calculate/calculate_base.py
johannes-moegerle: yes maybe, but since this is anyway rather fast (and also more code in python and not just C++) having it threaded and not multiprocessed seems to work ok for now. I currently think the overhead to define a separate Parameters and Results for the lifetimes is not yet worth it.

src/pairinteraction_gui/app.py
johannes-moegerle: yes they could, since the Application instance is a global instance we could move this simply to the application class, however I am not completely sure if this makes sense conceptually.

src/pairinteraction_gui/page/base_page.py
seweber: I think my confusion came from using `[...]_<type>` vs. `<type>_[...]`. I wonder whether  `caclulate_<type>`, `page_<type>`, and `config_<type>` are viable? If there is an analogicity between calculate, page, and config, using the same naming scheme would emphasize this analogicity.

src/pairinteraction_gui/app.py
seweber: Can one say that ALL_PROCESSES and  ALL_THREADS should have the same lifetime as Application? If yes, this could be an argument that these variables should really be member variables of Application. But I am not sure either.

johannes-moegerle: I agree, moved it to the Application class

src/pairinteraction_gui/main_window.py
johannes-moegerle: actually overriding close was correct here, I changed it to override closeEvent and made sure the normal closEvent is also triggered

src/pairinteraction_gui/calculate/calculate_base.py
johannes-moegerle: added a comment to calculate method of the lifetimes page

seweber (approved):

PR #231, opened 2025-04-08
Add state class

johannes-moegerle:


docs/tutorials/examples_python/concepts.ipynb
```diff
@@ -243,12 +243,12 @@
     "eigenstate_number = 2\n",
     "print(f\"The coefficients of the eigenstate with index {eigenstate_number} are:\")\n",
     "print(f\"{eigenbasis.get_coefficients().toarray()[:, eigenstate_number]}\")\n",
-    "print(f\"It has the largest overlap with the ket: {eigenbasis.get_corresponding_ket(eigenstate_number)}\")\n",
+    "print(f\"It has the largest overlap with the ket: {eigenbasis.kets[eigenstate_number]}\")\n",
```
seweber, line 246: ket[ket_idx] differs from ket[state_idx], doesn't it?

src/pairinteraction/_wrapped/basis/basis.py
```diff
@@ -78,26 +81,15 @@ def get_coefficients(self) -> "csr_matrix":
...
-        else:
-            cpp_basis = self._cpp.get_corresponding_state(ket_or_index._cpp)  # type: ignore [arg-type]
-        return type(self)._from_cpp_object(cpp_basis)
-
-    def get_corresponding_state_index(self, ket_or_index: Union[KetType, int]) -> int:
-        if isinstance(ket_or_index, (int, np.integer)):
-            return self._cpp.get_corresponding_state_index(ket_or_index)
-        return self._cpp.get_corresponding_state_index(ket_or_index._cpp)  # type: ignore [arg-type]
-
-    def get_corresponding_ket(self: "Self", state_or_index: Union["Self", int]) -> KetType:
```
seweber, line 93: I do not yet completely understand why the corresponding_ket methods are no longer needed with the new design.

src/pairinteraction/_wrapped/basis/basis_atom.py
```diff
@@ -134,35 +137,54 @@ def __repr__(self) -> str:
...
         return f"{type(self).__name__}({args})"
 
+    @property
+    def database(self) -> Database:
+        """The database used for this object."""
+        return self._database
+
     @overload
-    def get_amplitudes(self, ket_or_basis: KetAtom) -> "NDArray": ...
+    def get_amplitudes(self, other: KetAtom) -> "NDArray": ...
```
seweber, line 146: "other" sounds as if the object would be of same type as self

tests/test_matrix_elements.py
```diff
@@ -36,7 +36,7 @@ def test_electric_dipole_matrix_element() -> None:
     ket_final = pi.KetAtom("Rb", n=60, l=1, j=0.5, m=0.5)
 
     dipole = ket_initial.get_matrix_element(ket_final, "electric_dipole", 0, unit="e a0")
-    assert dipole > 0
+    assert abs(dipole) > 0
```
seweber, line 39: Should we really not check the sign (also bellow)?

src/pairinteraction/_wrapped/state/state_atom.py
seweber, line 1: Just to ensure that I understand it correctly: StateAtom is on the same level as BasisAtom (both wrap the BasisAtom cpp class) and thus both belong into _wrapped?

src/pairinteraction/_wrapped/state/state_pair.py
seweber, line 1: Why does this class does not have the methods that StateAtom has (if it is work in progress, I would add the method headers and raise a not yet implementer error)? Shouldn't the StateBase class enforce that for StateAtom and State pair, the same methods are present?

src/pairinteraction/_wrapped/state/state.py
seweber, line 1: Cross check that similar methods as in the Basis class are resent. If methods are missing, I would already add the method headers and raise a not yet implemented error.

src/pairinteraction/_wrapped/state/state.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_coefficients(self) -> "NDArray":
+        """Return the coefficients of the state as 1d-array."""
+        return self._basis.get_coefficients().toarray().flatten()
+
+    def get_corresponding_ket(self) -> KetType:
+        cpp_ket = self._basis._cpp.get_corresponding_ket(self._basis._cpp)  # type: ignore [arg-type]
+        return self._TypeKet._from_cpp_object(cpp_ket)
+
+    def get_corresponding_ket_index(self) -> int:
```
seweber, line 88: Hmm, here are corresponding_ket methods but not in the Basis class. Probably, I overlooked something.

tests/test_matrix_elements.py
johannes-moegerle: ah true, will change it to `.real` and also .complex == 0

src/pairinteraction/_wrapped/state/state.py
johannes-moegerle: added abstractmethods

src/pairinteraction/_wrapped/state/state_pair.py
johannes-moegerle: yes, added abstractmethods in StateBase and raise NotImplemented in StatePair

src/pairinteraction/_wrapped/basis/basis.py
johannes-moegerle: actually `get_corresponding_ket(state)` is not working due to not implemented in the c++ backend,  `get_corresponding_ket(state_idx)` would work, but we dont want this on the long run (but rather want `get_corresponding_ket(basis.states[state_idx])` which however currently is not working). Should we thus provide this for now?

seweber: I would add get_corresponding_ket(state) and raise a not yet implemented error

seweber (approved):

PR #237, opened 2025-04-24
A few small fixes

johannes-moegerle:


.github/workflows/documentation.yml
```diff
@@ -122,7 +122,10 @@ jobs:
     - name: Install pairinteraction into virtual environment
       run: uv pip install -e .[tests]  # -e is needed for coverage to work
     - name: Run pytest with coverage
-      run: uv run --no-project coverage run --source=src,tests -m pytest
+      run: |  # also run with --generate-reference to get 100% tests coverage and test the reference generation
+        uv run --no-project coverage run --source=src,tests -m pytest
+        uv run --no-project coverage run --append --source=src,tests -m pytest --generate-reference
+        uv run --no-project coverage run --append --source=src,tests -m pytest
```
seweber, line 128: What is the difference between the first call to pytest and the third call to pytest?

seweber (approved):

.github/workflows/documentation.yml
johannes-moegerle: its run against the new generated reference data, in this way also the generation of the reference data is tested

PR #238, opened 2025-04-25
Improve gui

johannes-moegerle:
# GUI changes
- added more state annotations (as filled diamonds) to the gui plot
- added pair m range to gui basis pair
- improve layout of plot window
- add Calculation config tab
- add better navigation toolbar to plot

# Implementation changes
- move all style sheets into separate folder theme
- improve implementation of Item objects
- added tests for gui calculate and compare them to the starkmap and pair potential reference data

# Future PR
- rename gui files and classes to match pattern like "ConfigBasisOneAtom", ... (extra PR)
- loading gif for downloading database 
- fix bug on windows


src/pairinteraction_gui/calculate/calculate_two_atoms.py
```diff
@@ -25,20 +25,25 @@
 class ParametersTwoAtoms(Parameters["TwoAtomsPage"]):
     """Parameters for the two atoms calculation."""
 
-    pair_basis_energy_delta: float = 0
+    pair_delta_energy: float = 999
```
seweber, line 28: Can you initialize to None? In case somebody does not like to restrict the pair_delta_energy for small quantum numbers n, setting it by default to 999 might lead to unexpected behavior.

src/pairinteraction_gui/config/base_config.py
```diff
@@ -18,3 +18,6 @@ def __init__(self, parent: "SimulationPage") -> None:
         """Initialize the base section."""
         self.page = parent
         super().__init__(parent)
+
+    def postSetupWidget(self) -> None:
```
seweber, line 22: Is there really no better place to add a stretch? Because this looks like a rather common operation to me, I would have expected that this does not have to be done in some postSetupWidget function; but it might well be that I am wrong.

src/pairinteraction_gui/config/basis_config.py
```diff
@@ -193,50 +184,47 @@ class RestrictionsBase(WidgetV):
...
+            self, "Δf", vdefault=5, tooltip="Restriction for the Total angular momentum f", checked=False
+        )
+        self.items["m"] = QnItemInt(
+            self, "Δm", vdefault=5, tooltip="Restriction for the Magnetic quantum number m", checked=False
+        )
 
         # TODO this is not yet implemented
-        # spin_boxes["Δl_ryd"] = DoubleSpinBox(
-        #     self, vdefault=3, tooltip="Restriction for the Orbital angular momentum l_ryd of the Rydberg electron"
+        # self.items["l_ryd"] = QnItemDouble(
```
seweber, line 224: Can this commented-out code be transformed into an issue?

src/pairinteraction_gui/page/base_page.py
```diff
@@ -115,54 +78,51 @@ class CalculationPage(SimulationPage):
...
+        bottom_layout.addWidget(export_button, stretch=1)
 
-        self.layout().addLayout(control_layout)
+        self.layout().addLayout(bottom_layout)
+
+        # Setup loading animation
+        self.loading_label = QLabel(self)
+        gif_path = Path(__file__).parent.parent / "images" / "loading.gif"
+        self.loading_movie = QMovie(str(gif_path))
+        self.loading_movie.setScaledSize(QSize(100, 100))  # Make the gif larger
```
seweber, line 122: If the GIF is increased in size, is the quality still fine? Otherwise, I would use its original size.

src/pairinteraction_gui/page/base_page.py
```diff
@@ -216,8 +176,7 @@ def update_plot(self, parameters: Parameters[Any], results: Results) -> None:
 
         self.plotwidget.plot(x_values, energies, overlaps, x_label)
 
-        ind = 0 if parameters.n_atoms == 1 else -1
-        self.plotwidget.add_cursor(x_values[ind], energies[ind], results.state_labels_0)
+        self.plotwidget.add_cursor(x_values, energies, results.state_labels)
```
seweber, line 179: Why does the function take state_labels and not state_labels[0]? At some other places state_labels_0 was replaced by state_labels[0].

src/pairinteraction_gui/config/ket_config.py
```diff
@@ -103,13 +82,13 @@ def setupOneKetAtom(self) -> None:
         stacked_qn.addNamedWidget(QnMQDT(m_is_int=False), "mqdt_halfint")
 
         for _, widget in stacked_qn.items():
-            for item in widget.items:
+            for _, item in widget.items.items():
```
seweber, line 85: Wouldn't values() be more suitable than items()? Also at other places.

src/pairinteraction_gui/config/ket_config.py
```diff
@@ -285,48 +258,44 @@ class QnSQDT(QnBase):
...
+        self.items = {}
         super().__init__(parent)
 
     def setupWidget(self) -> None:
-        spin_boxes = self._spin_boxes = {}
-
-        spin_boxes["n"] = IntSpinBox(self, vmin=1, vdefault=80, tooltip="Principal quantum number n")
-
-        spin_boxes["l"] = IntSpinBox(self, tooltip="Orbital angular momentum l")
+        self.items["n"] = QnItemInt(self, "n", vmin=1, vdefault=80, tooltip="Principal quantum number n")
```
seweber, line 265: How is vmax dealt with? Ideally, it takes into account the database. Also, if some delta is chosen where the user would expect that states are added to the basis that are not included in the database, at least a warning should be thrown. This could be a topic for another PR/issue.

src/pairinteraction_gui/config/ket_config.py
```diff
@@ -285,48 +258,44 @@ class QnSQDT(QnBase):
...
+        self.items["l"] = QnItemInt(self, "l", tooltip="Orbital angular momentum l")
 
         s = self.s
         if s % 1 == 0:
-            spin_boxes["j"] = IntSpinBox(self, vmin=int(s), vdefault=int(s), tooltip="Total angular momentum j")
-            spin_boxes["m"] = IntSpinBox(self, vmin=-999, vdefault=0, tooltip="Magnetic quantum number m")
+            self.items["j"] = QnItemInt(self, "j", vmin=int(s), vdefault=int(s), tooltip="Total angular momentum j")
+            self.items["m"] = QnItemInt(self, "m", vmin=-999, vdefault=0, tooltip="Magnetic quantum number m")
```
seweber, line 271: It looks strange to only set vmin as m is completely symmetric around zero. Maybe, we could avoid setting min/max altogether? Also at other places. Or is it necessary to configure min/max for the spin boxes?

src/pairinteraction_gui/config/ket_config.py
```diff
@@ -285,48 +258,44 @@ class QnSQDT(QnBase):
...
         )
-        spin_boxes["s"] = DoubleSpinBox(self, vmin=0, vmax=1, vstep=0.1, tooltip="Spin s")
-        spin_boxes["j"] = DoubleSpinBox(self, vstep=1, tooltip="Total angular momentum j")
+        self.items["s"] = QnItemDouble(self, "s", vmin=0, vmax=1, vstep=0.1, tooltip="Spin s")
+        self.items["j"] = QnItemDouble(self, "j", vstep=1, tooltip="Total angular momentum j")
 
         if self.m_is_int:
-            spin_boxes["f"] = IntSpinBox(self, vmin=0, vdefault=0, tooltip="Total angular momentum f")
-            spin_boxes["m"] = IntSpinBox(self, vmin=-999, vdefault=0, tooltip="Magnetic quantum number m")
+            self.items["f"] = QnItemInt(self, "f", vmin=0, vdefault=0, tooltip="Total angular momentum f")
```
seweber, line 293: I would find "QnSpinBoxInt" or "SpinBoxQnInt" more descriptive because such a name would tell me immediately that this is a spin box object in the GUI. But I do not have a strong feeling. If item fits better to the names of other objects, I would also be fine with item.

seweber: Ok, I just saw that you also have a "RangeItem" and items are in general more than just a SpinBoxes.

src/pairinteraction_gui/plotwidget/navigation_toolbar.py
```diff
@@ -0,0 +1,33 @@
...
+    See Also:
+    https://stackoverflow.com/questions/12695678/how-to-modify-the-navigation-toolbar-easily-in-a-matplotlib-figure-window/15549675#15549675
+
+    """
+
+    toolitems = (
+        ("Zoom", "Zoom to rectangle\nx/y fixes axis", "zoom_to_rect", "zoom"),
+        ("Pan", "Left button pans, Right button zooms\nx/y fixes axis, CTRL fixes aspect", "move", "pan"),
+        ("Home", "Reset original view", "home", "home"),
+        # (None, None, None, None),
```
seweber, line 23: I would remove the commented-out code. If one likes to add additional items, the link is enough as a reference from my point of view.

tests/test_starkmap.py
```diff
@@ -47,11 +47,27 @@ def test_starkmap(generate_reference: bool) -> None:
         np.savetxt(reference_overlaps_file, overlaps)
         pytest.skip("Reference data generated, skipping comparison test")
 
-    np.testing.assert_equal(kets, np.loadtxt(reference_kets_file, dtype=str, delimiter="\t"))
+    compare_starkmap_to_reference(eigenenergies, overlaps, eigenvectors, kets)
+
+
+def compare_starkmap_to_reference(
```
seweber, line 53: Looks very similar to compare_pair_potential_to_reference. Can this be abstracted away into a separate file, similarly as it is done with the constants?

tests/ruff.toml
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2024 Pairinteraction Developers
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+extend = "../pyproject.toml"
+lint.extend-ignore = ["TID252"]  # inside tests i would prefer relative imports, since absolute imports from tests looks weird
```
seweber, line 5: However, we can have a mixture of relative and absolute imports, and autoformatting does not work.

tests/test_gui.py
```diff
@@ -4,12 +4,18 @@
 import time
 from typing import TYPE_CHECKING
 
+import numpy as np
+from pairinteraction_gui.calculate.calculate_one_atom import ParametersOneAtom, _calculate_one_atom
```
seweber, line 8: At first glance, it looks strange to me that an underscore function has to be imported for the test case. However, it could be that I missed something and that having the function start with an underscore makes sense.

src/pairinteraction_gui/qobjects/item.py
```diff
@@ -109,8 +171,8 @@ def __init__(
         self,
         parent: QWidget,
         label: str,
-        value_range: tuple[float, float] = (0, 999),
-        value_defaults: tuple[float, float] = (0, 0),
+        vdefaults: tuple[float, float] = (0, 0),
+        vrange: tuple[float, float] = (-999, 999),
```
seweber, line 175: I wonder whether `vranges = ((-999, 999), (-999, 999))` could make sense to be more consistent with the definition of `vdefaults`?

seweber (approved):

src/pairinteraction_gui/calculate/calculate_two_atoms.py
johannes-moegerle: I added a checkbox to make it selectable and made the default np.inf.
In addition I added some logic to handle np.inf as min/max energy range in `basis_pair.py`, if you have better suggestions for that please let me know

src/pairinteraction_gui/config/base_config.py
johannes-moegerle: One also could do it for every config individually.
But since for also other reasons I added the postSetupWidget functionality I thought it is nice to be able to also use it here

src/pairinteraction_gui/config/basis_config.py
johannes-moegerle: there is already an issue on atom-pairinteraction: https://github.com/atom-pairinteraction/pairinteraction/issues/177
at some point we should probably clean up the issues and move the remaining once to this repo

johannes-moegerle: removed the commented out code for now

src/pairinteraction_gui/page/base_page.py
johannes-moegerle: actually it is made smaller, so yes quality is still fine

src/pairinteraction_gui/page/base_page.py
johannes-moegerle: because add_cursor now also adds all state labels and not only for the first or last parameter

src/pairinteraction_gui/config/ket_config.py
johannes-moegerle: agree

src/pairinteraction_gui/config/ket_config.py
johannes-moegerle: The default currently is 999 and not taking any database into account.
Throwing a warning if states should be added but cannot is probably a generic issue (and might be best addressed in the cpp code?
I opend a issue for that: https://github.com/pairinteraction/pairinteraction/issues/250

src/pairinteraction_gui/config/ket_config.py
johannes-moegerle: I does not make sense to allow negative values for the quantum numbers n,l,j, so I definitely would set a min value for them.
Also it is necessary to configure a max value, at least not configuring any value results in the limits (0,99)

src/pairinteraction_gui/plotwidget/navigation_toolbar.py
johannes-moegerle: agree

tests/test_starkmap.py
johannes-moegerle: maybe in the future, if we add more similar tests

tests/ruff.toml
johannes-moegerle: how do you mean? Now pairinteraction is treated as third party library and not the current package (which it is not) so I even prefer how it is now sorted

tests/test_gui.py
johannes-moegerle: for this PR it is because the underscore function is the one without the added multiprocessing (`run_in_other_process`) which here does not yet work, in https://github.com/pairinteraction/pairinteraction/pull/245 we then test both the raw udnerscore function and the multiprocessed function

src/pairinteraction_gui/qobjects/item.py
johannes-moegerle: since both describe the same quantity they always should have the exact same range limits, so for me this would be only unnecessary duplication

hmenke:
I wrote this in the Mattermost chat already, but I noticed that the input boxes for the external fields max out at 999.0. Especially for the magnetic fields this is very little, since 1000 Gauss is only 0.1 Tesla. At this strength you barely see anything. There is almost no state mixing for the example `|Rb:80,S_1/2,1/2>` configuration.

johannes-moegerle:
> I wrote this in the Mattermost chat already, but I noticed that the input boxes for the external fields max out at 999.0. Especially for the magnetic fields this is very little, since 1000 Gauss is only 0.1 Tesla. At this strength you barely see anything. There is almost no state mixing for the example `|Rb:80,S_1/2,1/2>` configuration.

I just updated this, would 9999.9 be sufficient?

johannes-moegerle:
ok actually pyside allows to set maximum to inf, so probably we go with this

PR #241, opened 2025-04-29
Support user-defined green tensors

seweber:


johannes-moegerle (approved):
Looks good, about the bindings/wrapping we can maybe shortly discuss how the workflow of constructing and creating green tensors should look like, before I start creating the wrappings, but we can also outsource this to a different PR

Issue #244, opened 2025-05-07
Failing GUI calculation on Windows


Babalion:
We tried to install the new pairinteraction version via pip on Windows.
We then start the gui via `pairinteraction gui`.
When we start the calculation, an error occurs:

![Image](https://github.com/user-attachments/assets/91f586af-50a0-41d0-8b5c-663d926fede6)

johannes-moegerle:
Thanks for reporting this bug.
This should be fixed by #245, which was just merged.
There will be a new pypi release today or tomorrow, which you then can try out.

johannes-moegerle:
There is a new release on pypi, you can install it via
```
pip install "pairinteraction>=2.0.1"
```

Feel free to reopen this issue, or open a new issue if you run into another problem.

PR #245, opened 2025-05-09
Fix gui on windows

johannes-moegerle:
fixing #244

### DIFF
--- a/src/pairinteraction_gui/worker.py
+++ b/src/pairinteraction_gui/worker.py
@@ -2,13 +2,14 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
 import logging
+import os
 from functools import wraps
-from multiprocessing import Process, SimpleQueue
-from typing import TYPE_CHECKING, Any, Callable, TypeVar
+from multiprocessing.pool import Pool
+from threading import Thread
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Optional, TypeVar
 
 from PySide6.QtCore import QObject, QThread, Signal
-
-from pairinteraction_gui.app import Application
+from PySide6.QtWidgets import QApplication
 
 if TYPE_CHECKING:
     from typing_extensions import ParamSpec
@@ -27,13 +28,22 @@ class WorkerSignals(QObject):
     result = Signal(object)
 
 
-class Worker(QThread):
-    """Simple worker class to run a function in a separate thread."""
+class MultiThreadWorker(QThread):
+    """Simple worker class to run a function in a separate thread.
+
+    Example:
+    worker = Worker(my_function, arg1, arg2, kwarg1=value1)
+    worker.signals.result.connect(process_result)
+    worker.start()
+
+    """
+
+    all_threads: ClassVar[set["MultiThreadWorker"]] = set()
 
     def __init__(self, fn: Callable[..., Any], *args: Any, **kwargs: Any) -> None:
-        super().__init__(Application.instance())
+        super().__init__(QApplication.instance())
 
-        Application.all_threads.add(self)
+        self.all_threads.add(self)
 
         self.fn = fn
         self.args = args
@@ -44,7 +54,7 @@ class Worker(QThread):
 
     def run(self) -> None:
         """Initialise the runner function with passed args, kwargs."""
-        logger.debug("Starting thread %s", self)
+        logger.debug("Run on thread %s", self)
         success = False
         try:
             result = self.fn(*self.args, **self.kwargs)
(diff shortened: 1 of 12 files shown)


seweber:
I looked through the last commits that are in addition to the commits of #238 and just have a few questions:
1. Is `return sys.exit(app.exec())` really recommended? What does returning after calling exit do? Moreover, calling `sys.exit` within `if __name__ == "__main__"` could be a better practice than calling it within the main function? Intuitively, I would have expected the main function to return an exit code, but not call sys.exit itself.
2. Just to be sure that I understood it correctly: `all_processes = list(Application.all_processes)` makes a shallow copy, i.e., the entries within the list and the copied list are referring to the same process objects?
3. I also wonder why exactly a copy of all processes is needed. Is the problem that a process can terminate and then be removed from the list at other places in the code while we are looping over the list? If this causes an issue, I would expect that using some mutex or a different datatype for the list, would be a better solution than making a copy of the list.
4. "# -s for more output" --> "# -s for text output to the terminal"
5. Is `PAIRINTERACTION_TEST_DATABASE_DIR` really needed? We can already pass the database dir as a command-line argument if we call pytest. In general, I wonder whether there could be a cleaner solution for handling the test mode than the one currently implemented in `tests/conftest.py`. To me, it looks like we  have different mechanisms for different ways of calling the tests, and all mechanisms ultimately do the same. I wonder whether one could simplify things.

johannes-moegerle:
> 1. Is `return sys.exit(app.exec())` really recommended? What does returning after calling exit do? Moreover, calling `sys.exit` within `if __name__ == "__main__"` could be a better practice than calling it within the main function? Intuitively, I would have expected the main function to return an exit code, but not call sys.exit itself.

I am not really sure what the convention is, however, I found one [blog post](https://www.pythonguis.com/faq/clean-up-on-exit-application/) where there seems to be problems if you dont pass app.exec inside sys.exit.
Not really sure about it, but since so far the exit seemed to worked fine i would rather not change it if we dont have anything specific in mind we want to fix

johannes-moegerle:
> 2. Just to be sure that I understood it correctly: `all_processes = list(Application.all_processes)` makes a shallow copy, i.e., the entries within the list and the copied list are referring to the same process objects?

Yes exactly.


> 3. I also wonder why exactly a copy of all processes is needed. Is the problem that a process can terminate and then be removed from the list at other places in the code while we are looping over the list? If this causes an issue, I would expect that using some mutex or a different datatype for the list, would be a better solution than making a copy of the list.

If the process finishes it removes itself from that set, so yes the problem is it can happen, that the process is finishing up during the quit/exit is called.
Probably there is a better solution to that, but i think here the simple solution of just making a copy is sufficient

src/pairinteraction_gui/worker.py
```diff
@@ -88,17 +88,32 @@ def terminate_all_threads(cls) -> None:
...
+
         self.fn_name = fn_name
         self.args = args
         self.kwargs = kwargs
 
-        self.queue: SimpleQueue[Any] = SimpleQueue()
+    @property
+    def pool(self) -> Pool:
+        if self._pool is None:
+            return self.create_pool()
```
seweber: "self._pool is None" cannot happen because it is created in the constructor

src/pairinteraction_gui/worker.py
```diff
@@ -88,17 +88,32 @@ def terminate_all_threads(cls) -> None:
...
+        return self._pool
+
+    @classmethod
+    def create_pool(cls, n_processes: int = 1) -> Pool:
```
seweber: Why is this class method needed? Wouldn't it be sufficient to add n_processes as an argument to the constructor?

src/pairinteraction_gui/worker.py
```diff
@@ -104,16 +106,46 @@ def __init__(self, fn_name: str, *args: Any, **kwargs: Any) -> None:
...
+                return
+            raise ValueError("Pool already created with different number of processes.")
+        if cls._async_worker is not None:
+            if not run_async:
+                logger.debug("Waiting for async worker to finish.")
+                cls._async_worker.join()
+                logger.debug("Async worker finished.")
+            return
+
+        if not run_async:
```
seweber, line 126: Is it beneficial not to use async in some scenarios? If this is not the case and we only use async, I would remove the non-async option.

src/pairinteraction_gui/worker.py
```diff
@@ -136,16 +168,21 @@ def run(self) -> None:
...
+
         if cls._pool is None:
             return
 
         cls._pool.terminate()
         cls._pool = None
-
         logger.debug("Process pool terminated.")
 
+        if create_new_pool_async:
```
seweber, line 183: create_new_pool_async -> create_new_pool. The relevant aspect of this parameter is whether a new pool should be created or not, and not whether it is async or not.

PR #247, opened 2025-05-09
Load MKL only if compiled with mkl

seweber:
Allows for installing a version of pairinteraction without MKL by installing the Python dependencies manually except for MKL, and then running ` pip install --no-deps .`.


hmenke:
Couldn't this be done entirely in Python? Also I wonder whether an extra CMake flag is necessary for this. Can't we just base this off `MKL_FOUND`?

johannes-moegerle (approved):

seweber:
I don't think that it can be done entirely in Python. The pyproject.toml has MKL as a dependency so that MKL_FOUND is always true on Windows/Linux. In theory, we could use MKL as an optional dependency. However, then we would have to provide two different versions of pairinteraction, one linking against MKL and one not. Having just one version that dynamically loads MKL if it is available, would probably require to not let Eigen use MKL. Thus, I don't see a good alternative to the current approach.

seweber:
One thing that we could do: Instead of having to set "DWITH_MKL_IF_AVAILABLE=OFF", one could require that a user builds pairinteraction using "--no-deps" and installs all dependencies except for MKL manually. Probably, this is the cleaner solution. If one does not like to use MKL, it does not make sense to install the standard dependencies that include MKL.

hmenke:
Yes, for Flatpak I build with `--no-deps`, because there I can't install `mkl` and `tbb` wheels. This is hopefully the only situation where MKL is not available.

Regarding doing this purely in Python, I was thinking of generating a pure Python file from CMake instead of yet another C++ module.

seweber:
Ah, I see. Generating a Python file from CMake should work as well, but I am not sure whether this would be a cleaner solution: For running mypy or if one performs an editable installation, it might be necessary that the Python file is generated within the source tree. In addition, for future extensions, an info struct in C++ might be more flexible than an autogenerated Python file.

hmenke (approved):

PR #252, opened 2025-05-14
Wrap green tensor

johannes-moegerle:


src/pairinteraction/complex.py
```diff
@@ -4,10 +4,13 @@
 from pairinteraction._wrapped import (
     BasisAtomComplex as BasisAtom,
     BasisPairComplex as BasisPair,
+    ConstantEntryComplex as ConstantEntry,
```
seweber, line 7: Is it necessary to provide ConstantEntry and OmegaDependentEntry to the user, as the user cannot construct these objects directly?

seweber (changes requested):

src/pairinteraction/complex.py
johannes-moegerle: it might be useful for typehints in the same way as there is also Parity in this file, but we can also discuss about only providing classes and functions that should be used by an user in this file

johannes-moegerle: simplified the imports

seweber (approved):

PR #254, opened 2025-05-14
Deprecate energy_range of diagonalize function and method

johannes-moegerle:


seweber (approved):

johannes-moegerle:
dismissed in favor of using energy_range and energy_range_unit
also make sure, that we always either have just the name "unit" if it is really obvious or use for a property "name" the keywordargument "name_unit" (see new PR)

PR #259, opened 2025-05-16
Perturbative: add create_system_for_perturbative

johannes-moegerle:
Implement minimal changes for better perturbative calculations.

In a future PR restructure perturbative calculations as class structure ...

# Old comments for reference
# Things to discuss
- do we want to allow passing `min_population_admixture` to `get_c6(...)`, etc.? (already now has 7! arguments)
- changed `check_for_resonances` to only log errors and dont raise Exceptions, also always use `required_overlap=0.9` but maybe allow the user to call the function separately with custom required_overlap
- do we actually want the names `get_c6`, ... Since these are non-trivial calculations I maybe prefer `calc_c6`, ...
- Instead of returning the Perturbed Eigenvectors as sparse matrix, we could also return a list of PairStates
- improve the example notebooks (should we only make examples for `get_c6(...)` or also keep the examples with `get_c6_from_system(...)`

# Related future PR
- #266
- allow passing tuples of (value, unit) instead of pint.Quantities (probably we can then always allow this; this would make the unit argument in e.g. `set_electric_field(value, unit)` redundant ...)

### DIFF
--- a/src/pairinteraction/perturbative/__init__.py
+++ b/src/pairinteraction/perturbative/__init__.py
@@ -2,9 +2,15 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
 from pairinteraction.perturbative.perturbative import (
+    create_system_for_perturbative,
     get_c3_from_system,
     get_c6_from_system,
     get_effective_hamiltonian_from_system,
 )
 
-__all__ = ["get_c3_from_system", "get_c6_from_system", "get_effective_hamiltonian_from_system"]
+__all__ = [
+    "create_system_for_perturbative",
+    "get_c3_from_system",
+    "get_c6_from_system",
+    "get_effective_hamiltonian_from_system",
+]
--- a/src/pairinteraction/perturbative/perturbative.py
+++ b/src/pairinteraction/perturbative/perturbative.py
@@ -9,16 +9,22 @@ import numpy as np
 from scipy import sparse
 from scipy.sparse import csr_matrix
 
-from pairinteraction.units import QuantityArray, QuantityScalar
+from pairinteraction import (
+    complex as pi_complex,
+    real as pi_real,
+)
+from pairinteraction.units import AtomicUnits, QuantityArray, QuantityScalar, ureg
 
 if TYPE_CHECKING:
     from scipy.sparse import csr_matrix
 
-    from pairinteraction import (
-        complex as pi_complex,
-        real as pi_real,
+    from pairinteraction._wrapped.ket.ket_atom import KetAtom  # noqa: F401  # required for sphinx for KetPairLike
+    from pairinteraction._wrapped.ket.ket_pair import (  # noqa: F401  # required for sphinx for KetPairLike
+        KetAtomTuple,
+        KetPairComplex,
+        KetPairLike,
+        KetPairReal,
     )
-    from pairinteraction._wrapped.ket.ket_pair import KetPairLike
     from pairinteraction.units import NDArray, PintArray, PintFloat
 
     SystemPair = Union[pi_real.SystemPair, pi_complex.SystemPair]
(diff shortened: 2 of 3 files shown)


src/pairinteraction/perturbative/create_system.py
```diff
@@ -0,0 +1,173 @@
...
+            basis.get_matrix_elements(ket, "electric_dipole", q=q).to_base_units().magnitude
+            for ket in kets
+            for q in [+1, -1, 0]
+        ]
+        max_dipoles_au.append(np.max(np.abs(electric_dipoles_au)))
+
+    pi.diagonalize(system_atoms)
+
+    # if no distance vector is given, we set delta_energy_au to very small and use the min_number_of_kets below
+    delta_energy_au = QuantityScalar.from_unit(0.1, "MHz", "energy").to_base_unit()
```
EduardJBraun, line 97: I see that this is computationally faster, but this is so small it typically will only take into account 1 pair state, especially for lower lying Rydberg, such that we always have to go into the loop later. I would rather take the previous approach that we take 1 near lying state, or use an estimate that scales with the principal quantum number, i.e. some species dependent level spacing calculated for low lying states, divided by n*^3.

src/pairinteraction/perturbative/create_system.py
```diff
@@ -0,0 +1,173 @@
...
+            system_atoms,
+            energy=(min(pair_energies_au) - delta_energy_au, max(pair_energies_au) + delta_energy_au),
+            energy_unit=str(BaseUnits["energy"]),
+        )
+
+    basis_pair = get_basis_pair(delta_energy_au)
+
+    # minimum number of kets in the pair basis, even if the estimate of delta energy would lead to a smaller basis
+    # this is to avoid misleading results when considering large distances
+    min_number_of_kets = 2_000
```
EduardJBraun, line 133: I would rather suggest to do something model space dependent. If I have a model space of only 1 ket, 2000 is quite large, however for 20 kets, might be needed. I would suggest min_number_of_kets = 10 * len(ket_tuples)

src/pairinteraction/perturbative/create_system.py
```diff
@@ -0,0 +1,173 @@
...
+        # make a bisect search to get a sensible basis size between: min_number_of_kets and 1.5 * min_number_of_kets
+        while delta_energy_au < 1:  # stop if delta_energy_au is 1 and the basis is still very small
+            basis_pair = get_basis_pair(delta_energy_au)
+            if basis_pair.number_of_kets < min_number_of_kets:
+                min_delta = delta_energy_au
+                if max_delta is None:
+                    delta_energy_au *= 2
+                else:
+                    delta_energy_au = (delta_energy_au + max_delta) / 2
+            elif basis_pair.number_of_kets > min_number_of_kets * 1.5:
```
EduardJBraun, line 154: Do we need this fine tuning of the increasment?

src/pairinteraction/perturbative/effective_hamiltonian.py
```diff
@@ -0,0 +1,331 @@
...
+            "or you can use the `get_c3(...)` or `get_c6(...)` functions."
+        )
+
+    model_inds = _get_model_inds(ket_tuples, system_pair)
+    h_au = system_pair.get_hamiltonian().to_base_units().magnitude  # Hamiltonian in atomic units
+    h_eff_au, eigvec_perturb = _calculate_perturbative_hamiltonian(
+        h_au, model_inds, perturbation_order, return_only_specified_order=return_only_specified_order
+    )
+
+    check_for_resonances(ket_tuples, system_pair, eigvec_perturb, required_overlap=0.9)
```
EduardJBraun, line 94: If the user is sure there are no resonances, should'nt we give him the chance to omit this step to speed up calculations?

src/pairinteraction/perturbative/effective_hamiltonian.py
```diff
@@ -0,0 +1,331 @@
...
+    for i, m_ind in enumerate(model_inds):
+        overlaps[i, :] /= sparse.linalg.norm(overlaps[i, :])
+        if overlaps[i, m_ind] >= required_overlap:
+            continue
+        logger.error(
+            "The state %s has resonances with the following states, which might lead to unexpected/wrong results. "
+            "Please consider adding them to your model space:",
+            system_pair.basis.kets[m_ind],
+        )
+        print_above_admixture = (1 - required_overlap) * 0.05
```
EduardJBraun, line 325: Why do we use the 0.05 threshold? I would rather use a second threshold the user can choose freely

PR #263, opened 2025-05-21
Various fixes

johannes-moegerle:


src/pairinteraction/units.py
```diff
@@ -183,6 +192,24 @@ def to_pint_or_unit(self, unit: Optional[str]) -> Union[ValueType, PlainQuantity
...
         return self.to_unit(unit)
 
+    @classmethod
+    def convert_user_to_au(
+        cls,
+        value: Union[PlainQuantity[ValueType], ValueType],  # type: ignore [type-var]
+        unit: Optional[str],
+        dimension: DimensionLike,
+    ) -> ValueType:
+        return cls.from_pint_or_unit(value, unit, dimension).to_base_unit()
```
seweber, line 202: "au" and "base_unit" are used for the same thing if I see it correctly. I would consider replacing "base_unit" with "au" everywhere.

tests/test_db.py
```diff
@@ -15,13 +16,15 @@
...
-
-def fetch_id(n: int, l: float, f: float, s: float, connection: duckdb.DuckDBPyConnection, table: str) -> int:
-    return connection.execute(
-        f"SELECT id FROM '{table}' WHERE n = {n} AND f = {f} ORDER BY (exp_s - {s})^2+(exp_l - {l})^2 LIMIT 1"
-    ).fetchone()[0]
+def fetch_id(
+    n: int, l: float, f: float, s: float, connection: duckdb.DuckDBPyConnection, table: Union[str, Path]
+) -> int:
+    result = connection.execute(
+        f"SELECT id FROM '{table}' WHERE n = ? AND f = ? ORDER BY (exp_s - ?)^2+(exp_l - ?)^2 LIMIT 1",  # noqa: S608
```
seweber, line 24: I am not sure whether I like parametrized queries here. It is anyway only a test case, does not process user input, and also with your changes, it is not yet fully parametrized (the '{table}' is still hardcoded).

seweber (changes requested):

src/pairinteraction/units.py
```diff
@@ -171,7 +174,13 @@ def to_unit(
     ) -> ValueType:
         """Return the value of the quantity in the given unit."""
         contexts = self.get_contexts(self.dimension)
-        return self._quantity.to(unit, *contexts).magnitude  # type: ignore [no-any-return] # also a problem with pint with sparse matrix
+        try:
+            return self._quantity.to(unit, *contexts).magnitude  # type: ignore [no-any-return] # also a problem with pint with sparse matrix
+        except pint.errors.DimensionalityError:
+            if "spectroscopy" in contexts:
```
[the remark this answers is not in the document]
johannes-moegerle: the normal conversion is done in `return self._quantity.to(unit, *contexts).magnitude`
the try except is only for spectroscopy, which allows conversion like "eV * bohr^3" -> "GHz * bohr^3", which pint cannot do by default.
if "spectroscopy" is not in contexts the error from the normal conversion will simply be raised

src/pairinteraction/units.py
johannes-moegerle: yes, this is true, at some point i was not sure, if we should be more general than au, but this does not make sense anymore, i agree, i will change this

tests/test_db.py
johannes-moegerle: i dont mind, so i changed it back

src/pairinteraction/units.py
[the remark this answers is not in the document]
johannes-moegerle: add comment

seweber (approved):

PR #269, opened 2025-06-06
Gui various fixes

johannes-moegerle:
- a few various gui fixes and improvements
- start adding tests for the gui export to python functionality
- improve ket (and basis) quantum number (restrictions) (now for mqdt we have nu, s, f, m and l_ryd as qn (where f is replaced by j if I=0))

### DIFF
--- a/src/pairinteraction_gui/calculate/calculate_base.py
+++ b/src/pairinteraction_gui/calculate/calculate_base.py
@@ -3,6 +3,7 @@
 
 import logging
 from abc import ABC
+from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar, Union
 
 import numpy as np
@@ -19,11 +20,11 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from pairinteraction.units import NDArray
+    from pairinteraction_gui.config.basis_config import QuantumNumberRestrictions
+    from pairinteraction_gui.config.ket_config import QuantumNumbers
     from pairinteraction_gui.page import OneAtomPage, TwoAtomsPage
 
 logger = logging.getLogger(__name__)
-
-# FIXME: having all kwargs dictionaries being Any is a hacky solution, it would be nice to use TypedDict in the future
 
 UnitFromRangeKey: dict[RangesKeys, str] = {
     "Ex": "V/cm",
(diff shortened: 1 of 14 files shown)


tests/compare_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    from pairinteraction.units import NDArray
+
+
+REFERENCE_PATHS = {
+    "stark_map": Path(__file__).parent.parent / "data" / "reference_stark_map",
+    "pair_potential": Path(__file__).parent.parent / "data" / "reference_pair_potential",
+}
+
+
+def compare_to_reference(
```
seweber, line 19: compare_to_reference -> compare_eigensystem_to_reference

tests/test_gui.py
```diff
@@ -71,61 +99,41 @@ def test_main_window_basic(qtbot: "QtBot", window_starkmap: "MainWindow") -> Non
...
+    energies = np.array(results.energies) + ket_energy_0
+    compare_to_reference(REFERENCE_PATHS[reference_name], energies)
 
-    calculation_config = two_atoms_page.calculation_config
-    calculation_config.steps.setValue(5)
-    system_config = two_atoms_page.system_config
-    system_config.distance.setValues(1, 5)
+    # Test export to Python code
+    python_code = page._create_python_code()
+    python_code = python_code.replace("plt.show()", "")  # HACK, otherwise it will block the test
```
seweber, line 133: What would happen if one were using the agg backend for the test? Would the hack be needed then?

tests/compare_utils.py
```diff
@@ -0,0 +1,41 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from pairinteraction.units import NDArray
+
+
+REFERENCE_PATHS = {
```
seweber, line 13: Can't these paths be different for different tests? I would tend to specify them directly within the test_[...].py files.

seweber (changes requested):

src/pairinteraction_gui/calculate/calculate_base.py
```diff
@@ -136,19 +138,13 @@ def get_quantum_numbers(self, atom: Optional[int] = None) -> dict[str, Any]:
...
-        for key, delta in self.quantum_number_deltas[atom].items():
-            if key in self.quantum_numbers[atom]:
-                qn_restrictions[key] = (
-                    self.quantum_numbers[atom][key] - delta,
-                    self.quantum_numbers[atom][key] + delta,
-                )
-            else:
-                raise ValueError(f"Quantum number delta {key} not found in quantum numbers.")
-        return qn_restrictions
+        """Return the quantum numbers for the given ket."""
```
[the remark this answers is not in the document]
johannes-moegerle: yes it does return exactly the same, i added better typing so this is clearer now

tests/test_gui.py
johannes-moegerle: using agg still freezes the pytest run (although I am not sure why).
The only alternative I found is using pytest monkeypatch, which however is basically the same hack ...

PR #272, opened 2025-06-23
Add a button to fit c3/c6 to GUI

Patschke:
I found that feature in 1.x pairinteraction useful. So I implemented a simple version for 2.x.

There is a simple heuristic trying to track the correct potential curves, but as the selected data points are highlighted the user will know if it fails to do so correctly. 

fit_button

### DIFF
--- a/src/pairinteraction_gui/page/base_page.py
+++ b/src/pairinteraction_gui/page/base_page.py
@@ -87,6 +87,7 @@ class CalculationPage(SimulationPage):
 
         # Control panel below the plot
         bottom_layout = QHBoxLayout()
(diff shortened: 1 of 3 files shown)


johannes-moegerle:
Hi,

thanks for implementing this feature again!
I rebased the code to the current master and added a few minor fixes.

From my side, we can merge it now into master, such that it is available in the next release.

Best regards,
Johannes

seweber:
We have now released version [v2.2.0](https://github.com/pairinteraction/pairinteraction/releases/tag/v2.2.0) of pairinteraction, which includes the button to fit c3/c6. Thanks for the nice work!

PR #287, opened 2025-08-06
Add multiple doc versions

johannes-moegerle:


.github/workflows/documentation.yml
```diff
@@ -31,11 +31,33 @@ jobs:
...
+
+  build_sphinx_polyversion:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    container:
+      image: ghcr.io/pairinteraction/pairinteraction-debian:docker
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
hmenke, line 44: Completely unrelated, but using `fetch-depth: 0` dates back to when we used Git within CMake to determine the current version by inspecting Git tags for which you need the entire history, but since we don't do that anymore, this can be removed which should also speed up the cloning slightly.

docs/conf.py
```diff
@@ -20,6 +20,14 @@
 
 language = "en"
 
+# -- sphinx-polyversion -----------------------------------------------------
+if os.getenv("POLYVERSION_DATA"):
```
hmenke, line 24: Is `POLYVERSION_DATA` documented? Where do we set this?

docs/poly.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+#: Mapping of revisions to changes in build parameters
+BUILDER = {
+    "v0.9.10": UpdatedCommandBuilder(
+        Path("doc/sphinx/"),
+        pre_cmd=["cp", "Placeholder.SOURCE_DIR/conf.py.in", "Placeholder.SOURCE_DIR/conf.py"],
+        cmd=["sphinx-build", "--color", "-a", "-v", "--keep-going", Placeholder.SOURCE_DIR, Placeholder.OUTPUT_DIR],
+    ),
+    "v2.0.0": SphinxBuilder(Path("docs/"), args=["-a", "-v", "-W", "--keep-going"]),
```
hmenke, line 89: This seems a bit arbitrary. What about v2.1.0 and v2.2.0?

docs/poly.py
hmenke, line 1: I think this file would benefit from exhaustive commentary, because sphinx_polyversion is not exactly well-documented itself.

docs/templates/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+    <head>
+        <title>Redirecting to master branch</title>
+        <meta charset="utf-8" />
+        <meta
+            http-equiv="refresh"
+            content="0; url=./{{ latest.name }}/index.html"
+        />
+        <link rel="canonical" href="./{{ latest.name }}/index.html" />
+    </head>
```
hmenke, line 12: If I remember correctly `<meta http-equiv="refresh" content="0; url=..." />` is a non-standard extension, although supported by most browsers.
```suggestion
        <noscript><meta
            http-equiv="refresh"
            content="0; url=./{{ latest.name }}/index.html"
        /></noscript>
        <link rel="canonical" href="./{{ latest.name }}/index.html" />
    </head>
    <body onload="window.location.replace('./{{ latest.name }}/index.html')">
        If you are not redirected automatically, click <a href="./{{ latest.name }}/index.html">this link</a>.
    </body>
```

.github/workflows/documentation.yml
johannes-moegerle: sphinx-polyversion also needs access to the entire git history, thats why i added it here again

docs/conf.py
johannes-moegerle: it is automatically set when `sphinx-polyversion poly.py` is run.
It is not properly documented anywhere, but load() will fail if it was not set, and this way we still can use the same conf.py to simply build the latest doc via make

docs/poly.py
johannes-moegerle: kind of true, will update this

hmenke (approved):

PR #311, opened 2025-09-11
Small fixes

johannes-moegerle:


src/pairinteraction_gui/config/basis_config.py
```diff
@@ -274,7 +274,7 @@ def setupWidget(self) -> None:
         else:
             key = "f"
             description = "Restriction for the  total angular momentum f"
-        self.items[key] = QnItemDouble(self, "Δ" + key, vdefault=3, tooltip=description)
+        self.items[key] = QnItemInt(self, "Δ" + key, vdefault=3, tooltip=description)
```
seweber, line 277: Why "int" (also below)? Isn't it a half-int?

src/cpp/src/ket/KetPair.cpp
```diff
@@ -29,13 +29,18 @@ KetPair<Scalar>::KetPair(
 template <typename Scalar>
 std::string KetPair<Scalar>::get_label() const {
     std::string label;
-    std::string separator;
```
seweber, line 32: I would keep the separator trick, as we also use it in other parts of the C++ code.

seweber (changes requested):
Can we get the macOS build running? Maybe, substituting in the workflow

"brew reinstall $(brew --cache --bottle-tag=arm64_ventura spdlog nlohmann-json eigen tbb lapack fmt openssl gcc)"

by

"brew reinstall --force-bottle spdlog nlohmann-json eigen tbb lapack fmt openssl@3 gcc"

solves it? I hope that the previously fetched packages are used automatically without having to rely on the $(brew --cache ...) trick, which no longer seems to work.

src/pairinteraction_gui/config/basis_config.py
johannes-moegerle: f is half int, but this means meaningful Delta f is only int (if f = 2.5, we want to allow e.g. Delta f = 2 resulting in a range f=(0.5,4.5), having a Delta f = 0.5 does not make sense)

PR #315, opened 2025-09-19
Pytest use pairinteraction and pairinteraction.real

johannes-moegerle:


seweber (approved):
The docstrings at the top of the test files have been removed. In principle, I am OK with it, but I wanted to double-check whether this was intentional.

PR #386, opened 2026-04-07
Improve basis atom overlap

johannes-moegerle:
### TODO
- maybe the order of row, col for basis overlaps should be the other way around?


johannes-moegerle:
replaced by #419 and #423

PR #388, opened 2026-04-08
Various GUI fixes

johannes-moegerle:
- store all the systems and calculate the annotation labels just in time
- a lot of small fixes

# TODO in another PR
- allow editing fields even if they are not checked
- yrange and xrange in plotwindow setzen
- select which energy_0 defintion to use (with or without fields) for pair calculations
- add C6 page (after #391)

## see #393
- structure system parameters: (one checkbox for all efields, one for all b fields, put diamagnetism to b field, new title "Rydberg interaction" for distance, angle and multipole order)
- add ion interaction to gui

## see #396
- show diagonalization progress in gui below spinning wheel

## see #397
- clean up lifetime page (and add lifetime jupyter notebook)


seweber:
Looks cool :-) A few remarks on the labeling boxes:

* The boxes should be in the foreground, especially on top of the colorbar.
* Boxes should be closed if on clicks in an area with no curves.
* Instead of "+ -..." write "- ..."
* I like the waiting animation that appears when the labeling boxes are calculated. Consider to use this animation in a scaled up version also as waiting animation during calculation.
* If I click on the left half, labeling boxes should appear to the right. If I click on the right half, they should appear to the left (I would use this behavior for the curves and also the lifetimes). Otherwise, a part of the labeling boxes is invisible because it is outside of the canvas.

Remark on lifetimes panel:

*  The waiting animation shows only "Starting calculation..." which is missleading.
* The labeling boxes are centered on the bars but not where the mouse coursor is which feels strange.

docs/tutorials/examples_python/pair_potentials.ipynb
```diff
@@ -117,9 +117,9 @@
     "ax.set_xlabel(r\"Distance [$\\mu$m]\")\n",
     "ax.set_ylabel(\"Energy [GHz]\")\n",
     "\n",
-    "try:\n",
+    "if len({len(es) for es in eigenenergies}) <= 1:  # check if homogeneous shape\n",
```
seweber: For the tutorial, this code is a bit hard to understand. Maybe, being more explicit:

```
lengths = [len(es) for es in eigenenergies]
is_homogeneous = all(length == lengths[0] for length in lengths)
```

Or `np_eigenenergies = np.asarray(eigenenergies)` and then checking `np_eigenenergies.ndim == 2`?

src/pairinteraction_gui/calculate/calculate_base.py
```diff
@@ -254,14 +252,29 @@ def from_calculate(
...
 
-        MultiThreadWorker.task_checkpoint("Preparing state labels...")
-        steps_with_labels = [int(i) for i in np.linspace(0, parameters.steps - 1, parameters.number_state_labels)]
-        states_dict = {i: system_list[i].get_eigenbasis().states for i in steps_with_labels}
-        state_labels = {i: [s.get_label() for s in states] for i, states in states_dict.items()}
+        return cls(energies, energy_offset, ket_overlaps, list(system_list))
 
-        return cls(energies, energy_offset, ket_overlaps, state_labels)
+
+def diagonalize_with_progress(systems: Sequence[SystemBase[Any]], **diagonalize_kwargs: Any) -> None:
```
seweber: Now that we have the task_checkpoint functionality, I would not use the chunks approach (it can cost a lot of performance on a system with many CPU cores). Maybe, removing the progress indication from this PR and extending task_checkpoint in another PR to take, in addition to the status message, a boolean indicating whether the number of already diagonalized systems should be increased?

src/pairinteraction_gui/plotwidget/navigation_toolbar.py
```diff
@@ -28,3 +30,15 @@ class CustomNavigationToolbar(NavigationToolbar):
...
         super().__init__(canvas, parent, coordinates=False)
+        self._home_callbacks: list[Callable[[], None]] = []
+
+    def home(self, *args: Any) -> None:
+        """Reset view and notify registered callbacks."""
+        super().home(*args)
+        for cb in self._home_callbacks:
+            cb()
+
+    def refresh_home_view(self) -> None:
```
seweber: Somehow, I do not fully understand the role of home and refresh_home_view (what is home?). Can you expand the docstring?

src/pairinteraction_gui/page/base_page.py
```diff
@@ -175,10 +175,18 @@ def calculate(self) -> tuple[Parameters[Any], Results]:
         raise NotImplementedError("Subclasses must implement this method")
 
     def update_plot(self, parameters: Parameters[Any], results: Results) -> None:
+        self.plotwidget.canvas.draw()
         self.plotwidget.plot(parameters, results)
-        self.plotwidget.add_cursor(parameters, results)
+        self.plotwidget.setup_annotations(parameters, results)
+        self._plot_function(parameters, results)
         self.plotwidget.canvas.draw()
```
seweber, line 182: There is a lot self.plotwidget.canvas.draw(). In particular, there is a call to it before _plot_function, although its docstring says that _plot_function is called before canvas.draw(). Is the first canvas.draw() a workaround for avoiding display artifacts once a calculation is done? If yes, add a comment.

src/pairinteraction_gui/page/one_atom_page.py
```diff
@@ -71,6 +72,9 @@ def add_short_labels(
             used.add(short_label)
             self.plotwidget.canvas.ax.text(x_lim[0], energy, short_label, va="center", ha="right")
 
+        self.plotwidget.canvas.draw()
```
seweber: Why is here yet another canvas.draw() and refresh_home_view()?

src/pairinteraction_gui/plotwidget/plotwidget.py
```diff
@@ -140,37 +152,80 @@ def plot(self, parameters: Parameters[Any], results: Results) -> None:
...
+                va=va,
+                bbox={"boxstyle": "round,pad=0.5", "fc": "white", "alpha": 0.9, "ec": "gray"},
+                arrowprops={"arrowstyle": "->", "connectionstyle": "arc3", "color": "gray"},
+                clip_on=False,
+                annotation_clip=False,
+            )
+            ann.set_in_layout(False)
+            self._annotations[nearest] = ann
+            self.canvas.draw_idle()
+            show_status_tip(self, "Ready")
```
seweber: Not sure whether "Ready" is understandable from the context. If not, maybe, "State annotations calculated."?

seweber (changes requested):

docs/tutorials/examples_python/pair_potentials.ipynb
johannes-moegerle: `np.asarray(eigenenergies)` will already throw a ValueError, dont you think the comment is explicit enough?
the only other more explicit thing could be
`if all(len(es) == len(eigenenergies[0]) for es in eigenenergies):`

src/pairinteraction_gui/plotwidget/navigation_toolbar.py
johannes-moegerle: If a user clicks the home symbol in the navigation toolbar on the upper right in the plot this was previously resetting to the very first plot of the gui session.
refresh_home_view will set the current state as "home view" such that the home buttona always resets to the first view after the latest calculation.

Renamed to `reset_home_view` and added the following docstring:
```
"""Reset the home view to the current axes state.

I.e. if a user afterwards clicks the home/reset button, the view will be reset to the current view.
"""
```

src/pairinteraction_gui/page/base_page.py
johannes-moegerle: yes, added a comment

src/pairinteraction_gui/page/one_atom_page.py
johannes-moegerle: removed it

src/pairinteraction_gui/calculate/calculate_base.py
johannes-moegerle: agree, will open a new PR

seweber (approved):

PR #394, opened 2026-04-11
Make 'pairinteraction' launch the gui directly

seweber:


johannes-moegerle (approved):
Maybe we want to print the help message if a wrong argument is given?
This might help people who type `pairinteraction gui` which now gives the error:
```
usage: pairinteraction [-h] [--version] [--reload] [--log-level {DEBUG,INFO,WARNING,ERROR,CRITICAL}] {test,database,config} ...
pairinteraction: error: argument command: invalid choice: 'gui' (choose from 'test', 'database', 'config')
```
which might lead to confusion, whether there still is a gui.

Or specifically catch `pairinteraction gui` and print that they need to write just `pairinteraction` ?

johannes-moegerle:
The goal if this was to make `uvx pairinteraction` work, do you want to mention this in the docs?

PR #397, opened 2026-04-13
Update lifetime page

johannes-moegerle:


seweber (changes requested):
Looks good. Just two small details:
* If one clicks on the right side of the canvas to open a label box, the text within the label box should still be left-aligned.
* I would make the region smaller, in which the state with the highest overlap is selected for labeling. Currently, it is a bit confusing: if one clearly clicks on one point, the label can appear on another point.

seweber (approved):

PR #398, opened 2026-04-16
Improve state pair label

johannes-moegerle:
merge after #386


johannes-moegerle:
Minimal implementation is done in #411
Keep this PR as reference, for improving the C++ implemtation

johannes-moegerle:
replaced by #411 (and #423)

PR #399, opened 2026-04-16
Add a splash screen

seweber:


src/pairinteraction_gui/app.py
```diff
@@ -34,10 +35,10 @@ class Application(QApplication):
 
     signals = MainSignals()
 
-    @classmethod
-    def instance(cls) -> Application:
+    @staticmethod
+    def instance() -> Application:
         """Return the current instance of the application."""
-        return super().instance()  # type: ignore [return-value]
+        return QApplication.instance()  # type: ignore [return-value]
```
johannes-moegerle, line 41: why this change?

data/icon.tex
johannes-moegerle, line 1: this file is never used, was this on purpose because you wanted to keep it here for reference?

johannes-moegerle (changes requested):
the splash screen is always shown in the left top corner, can this be changed to appear more centered on the screen?

src/pairinteraction_gui/app.py
seweber: The linter in vscode was showing an error because the base class QApplication has this method as "@staticmethod" and not "@classmethod" (the method is inherited from QCoreApplication).

data/icon.tex
seweber: I wanted to make the file that defines the PI logo accessible in the current version of PairInteraction as well. Previously, one had to check out the legacy version to access it.

johannes-moegerle (approved):

PR #401, opened 2026-04-17
Implement permutation and inversion symmetry

seweber:


src/cpp/bindings/basis/Basis.py.cpp
```diff
@@ -176,7 +176,10 @@ static void declare_basis_pair_creator(nb::module_ &m, std::string const &type_n
         .def("add", &BasisPairCreator<T>::add)
         .def("restrict_energy", &BasisPairCreator<T>::restrict_energy)
         .def("restrict_quantum_number_m", &BasisPairCreator<T>::restrict_quantum_number_m)
-        .def("restrict_product_of_parities", &BasisPairCreator<T>::restrict_product_of_parities)
```
johannes-moegerle, line 179: why was restrict_product_of_parities removed?

src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -141,16 +171,107 @@ std::shared_ptr<const BasisPair<Scalar>> BasisPairCreator<Scalar>::create() cons
...
+                Eigen::Index column_index = state_index++;
+
+                // A pair state with identical one-atom states contributes with coefficient one
+                // if allowed by both symmetries.
+                transformation_triplets.emplace_back(row_index, column_index, 1);
+            } else if (id1 > id2) {
+                // Obtain the column index of the symmetrized state
+                std::array<size_t, 2> ordered_ids{std::max(id1, id2), std::min(id1, id2)};
+                auto [column_iterator, inserted] =
+                    ket_ids2state_index.try_emplace(ordered_ids, state_index);
```
johannes-moegerle, line 202: under which circumstances can/should this try_emplace silently fail?

johannes-moegerle: ah i see, probably because of id1 > id2 and id2 > id1.
Wouldnt it be more performant to only loop over id2 >= id1?

src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -141,16 +171,107 @@ std::shared_ptr<const BasisPair<Scalar>> BasisPairCreator<Scalar>::create() cons
...
+                                          transformation_triplets.end());
+
+    Eigen::Matrix<real_t, Eigen::Dynamic, 1> sum_of_squared_coefficients =
+        transformation_matrix.cwiseAbs2().transpose() *
+        Eigen::Matrix<real_t, Eigen::Dynamic, 1>::Ones(transformation_matrix.rows());
+    for (Eigen::Index column_index = 0; column_index < sum_of_squared_coefficients.size();
+         ++column_index) {
+        if (std::abs(sum_of_squared_coefficients[column_index] - 1) > numerical_precision) {
+            throw std::invalid_argument(
+                "The basis could not be symmetrized. This likely means that the specified parity "
```
johannes-moegerle, line 267: this check could then also be don (maybe even as assert) inside the loop if only looping over id1>=id2, since then the loop basically loops over the new states

src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -141,16 +171,107 @@ std::shared_ptr<const BasisPair<Scalar>> BasisPairCreator<Scalar>::create() cons
...
+                "restrictions are invalid for the given one-atom systems.");
+        }
+    }
+
+    // TODO: on the long run, construct the coefficient matrix directly
```
johannes-moegerle, line 272: again i wonder if this becomes trivial if you adapt the loop

src/cpp/src/basis/BasisPairCreator.test.cpp
```diff
@@ -16,16 +16,110 @@
...
 namespace pairinteraction {
 
 constexpr double HARTREE_IN_GHZ = 6579683.920501762;
 constexpr double VOLT_PER_CM_IN_ATOMIC_UNITS = 1 / 5.14220675112e9;
 constexpr double UM_IN_ATOMIC_UNITS = 1 / 5.29177210544e-5;
 
+namespace {
+template <typename Scalar>
+Eigen::SparseMatrix<Scalar, Eigen::RowMajor>
+build_manual_symmetrizer(const std::shared_ptr<const BasisPair<Scalar>> &basis,
```
johannes-moegerle, line 39: is this a useful test to just reimplement the symmetrizer?

src/cpp/src/system/System.cpp
```diff
@@ -285,7 +285,7 @@ System<Derived> &System<Derived>::diagonalize(const DiagonalizerInterface<scalar
         num_cols += matrix.cols();
     }
 
-    assert(static_cast<size_t>(num_rows) == basis->get_number_of_kets());
```
johannes-moegerle, line 288: why is both rows and columns number of states?
also why is cols only <= number of states?

tools/benchmarking/src/benchmarking/_run_arc.py
```diff
@@ -35,17 +35,17 @@ def main() -> None:
...
+    number_of_eigen_states = 50
     with timer() as get_duration:
-        calc.diagonalise(distances, number_of_states, progressOutput=False)
+        calc.diagonalise(distances, number_of_eigen_states, progressOutput=False)
     duration_diagonalization = get_duration()
 
     output = {
-        "number_of_kets": number_of_kets,
-        "number_of_states": number_of_states,
+        "number_of_basis_states": number_of_basis_states,
```
johannes-moegerle, line 47: why the change in naming convention for ket to basis states?

tests/test_pair_potential.py
johannes-moegerle, line 1: why did this file change?

src/pairinteraction_gui/calculate/calculate_two_atoms.py
johannes-moegerle, line 1: I think i would prefer to split the gui implementation in a separate PR

johannes-moegerle (changes requested):

johannes-moegerle:
first implement just the symmetry in the api, this is done in #407
then in another PR add it to the GUI

PR #411, opened 2026-05-08
Minimal improve state pair label

johannes-moegerle:


src/cpp/include/pairinteraction/basis/BasisAtom.hpp
```diff
@@ -56,6 +56,8 @@ class BasisAtom : public Basis<BasisAtom<Scalar>>,
 
     int get_ket_index_from_id(size_t ket_id) const;
 
+    std::shared_ptr<const Type> get_canonical_basis() const;
```
seweber: I would implement it in the base class and name the method `canonicalized`, analogously to `transformed`.

src/cpp/src/basis/BasisAtom.cpp
```diff
@@ -42,6 +42,12 @@ const std::string &BasisAtom<Scalar>::get_id_of_kets() const {
     return id_of_kets;
 }
 
+template <typename Scalar>
+std::shared_ptr<const BasisAtom<Scalar>> BasisAtom<Scalar>::get_canonical_basis() const {
+    return std::make_shared<BasisAtom<Scalar>>(Private{}, ketvec_t(this->kets),
```
seweber: Can't `this->kets` and `id_of_kets` be passed directly?

src/pairinteraction/state/state_pair.py
```diff
@@ -47,6 +52,80 @@ def __init__(self, ket: KetPair, basis: Any) -> None:
...
+            basis_atom_class = BasisAtomReal
+            system_atom_class = SystemAtomReal
+            basis_pair_class = BasisPairReal
+
+        basis_atoms_cpp = [self._cpp.get_basis1(), self._cpp.get_basis2()]
+        basis_atoms = [
+            basis_atom_class._from_cpp_object(basis_atom_cpp.get_canonical_basis())
+            for basis_atom_cpp in basis_atoms_cpp
+        ]
+        system_atoms = [system_atom_class(basis_atom).diagonalize() for basis_atom in basis_atoms]
```
seweber: To me, it looks strange that a SystemAtom needs to be diagonalized in the `get_label` method (this system should be diagonal by construction). Is the call to `.diagonalize()` really necessary?

src/pairinteraction/state/state_pair.py
```diff
@@ -47,6 +52,80 @@ def __init__(self, ket: KetPair, basis: Any) -> None:
...
+        # heuristic to quickly find a basis, which includes the max_kets most contributing kets
+        for number_of_kets in [1_000, 10_000]:
+            canonical_basis_pair = basis_pair_class.from_kets(
+                ket_pair, system_atoms, number_of_kets=number_of_kets, warn_number_of_kets=False
+            )
+
+            amplitudes = canonical_basis_pair.get_matrix_elements(self, ("identity", "identity"), (0, 0)).m
+            overlaps = np.abs(amplitudes) ** 2
+
+            _max_kets = min(max_kets, len(overlaps))
```
seweber: Add a comment explaining how this approach works to check whether the heuristic was successful.

src/pairinteraction/state/state_pair.py
```diff
@@ -47,6 +52,80 @@ def __init__(self, ket: KetPair, basis: Any) -> None:
...
+            largest_inds = np.argpartition(overlaps, -_max_kets)[-_max_kets:]
+            largest_inds = largest_inds[np.argsort(overlaps[largest_inds])[::-1]]
+
+            if self.norm**2 - np.sum(overlaps) <= overlaps[largest_inds[-1]]:
```
seweber: Add a comment motivating this criterion.

src/pairinteraction/state/state_pair.py
```diff
@@ -47,6 +52,80 @@ def __init__(self, ket: KetPair, basis: Any) -> None:
...
+        ]
+        system_atoms = [system_atom_class(basis_atom).diagonalize() for basis_atom in basis_atoms]
+
+        coeffs = np.abs(self.get_coefficients())
+        ket_pair = self.get_ket(int(np.argmax(coeffs)))
+
+        # heuristic to quickly find a basis, which includes the max_kets most contributing kets
+        for number_of_kets in [1_000, 10_000]:
+            canonical_basis_pair = basis_pair_class.from_kets(
+                ket_pair, system_atoms, number_of_kets=number_of_kets, warn_number_of_kets=False
```
seweber: Can we make system_atoms optional? I.e., does ket_pair come with enough information to trivially infer system_atoms from it?

src/pairinteraction/state/state_pair.py
```diff
@@ -47,6 +52,80 @@ def __init__(self, ket: KetPair, basis: Any) -> None:
...
+            basis_atom_class._from_cpp_object(basis_atom_cpp.get_canonical_basis())
+            for basis_atom_cpp in basis_atoms_cpp
+        ]
+        system_atoms = [system_atom_class(basis_atom).diagonalize() for basis_atom in basis_atoms]
+
+        coeffs = np.abs(self.get_coefficients())
+        ket_pair = self.get_ket(int(np.argmax(coeffs)))
+
+        # heuristic to quickly find a basis, which includes the max_kets most contributing kets
+        for number_of_kets in [1_000, 10_000]:
```
seweber: Instead of hardcoding numbers for the heuristic, can we let the user specify them? Maybe instead via number_of_kets, with an energy cutoff?

seweber (changes requested):

src/pairinteraction/state/state_pair.py
johannes-moegerle: I was not sure, if systems are diagonal after construction is still valid always?
I thought you changed the construction behaviour at some point, such that this (in principle) is not always true anymore?

src/pairinteraction/state/state_pair.py
johannes-moegerle: In any way we need a useful default for the labels printed in the gui, and I really think here specifying the number of kets is the better option: it does work quite reliable and automatically choosing a energy cutoff is almost impossible

src/pairinteraction/state/state_pair.py
johannes-moegerle: i dont think so

src/cpp/src/basis/Basis.cpp
```diff
@@ -610,6 +610,47 @@ void Basis<Derived>::get_indices_of_blocks_without_checks(
...
+    std::iota(result->ket_index_to_state_index.begin(), result->ket_index_to_state_index.end(), 0);
+
+    result->state_index_to_quantum_number_f.resize(n);
+    result->state_index_to_quantum_number_m.resize(n);
+    result->state_index_to_parity.resize(n);
+    result->_has_quantum_number_f = true;
+    result->_has_quantum_number_m = true;
+    result->_has_parity = true;
+
+    for (size_t i = 0; i < n; ++i) {
```
johannes-moegerle, line 636: I am not sure, if we should be that accurate about _has_quantum_number_f etc, or just initialize it with all false, since this method (for now) is anyway just used for state pair get label ...

seweber (approved):

PR #419, opened 2026-05-18
remove get_amplitudes in C++

johannes-moegerle:
For now simply remove amplitudes and overlaps methods in C++.
Next PR, improve get_matrix_elements path to work also for different basis


src/pairinteraction/state/state_atom.py
```diff
@@ -191,14 +184,7 @@ def get_overlap(self, other: Self | KetAtom) -> float:
...
         """
-        if not self.is_normalized() or (isinstance(other, StateAtom) and not other.is_normalized()):
-            logger.warning("WARNING: get_overlap is called with a non-normalized state.")
-
-        if isinstance(other, KetAtom):
-            return np.array(self._cpp.get_overlaps(other._cpp))[0]  # type: ignore [no-any-return]
-        if isinstance(other, StateAtom):
-            return self._cpp.get_overlaps(other._cpp).toarray().ravel()[0]  # type: ignore [no-any-return]
-        raise TypeError(f"Incompatible types: {type(other)=}; {type(self)=}")
+        return abs(self.get_amplitude(other)) ** 2
```
seweber, line 187: I have a slight preference for calling get_matrix_element directly rather than nesting functions. However, if the nesting fits better with how things are done at other parts in the code or you have another reason to keep the nesting, I would also be fine with it.

seweber (approved):

src/pairinteraction/state/state_atom.py
johannes-moegerle: nesting makes it possible to only once handle all the if KetAtom if StateAtom, ... in get_matrix_elements, and therefore significantly reduces code duplication, especially in the Basis classes

PR #423, opened 2026-05-26
Improve Basis.get_... methods

johannes-moegerle:
- In C++ allow `get_matrix_elements` with different basis, which in python then also allows this for get_overlaps ...
- improve the python test coverage of all get_... methods


src/cpp/src/basis/BasisAtom.cpp
```diff
@@ -45,7 +46,13 @@ const std::string &BasisAtom<Scalar>::get_id_of_kets() const {
 template <typename Scalar>
 Eigen::VectorX<Scalar> BasisAtom<Scalar>::get_matrix_elements(std::shared_ptr<const ket_t> ket,
                                                               OperatorType type, int q) const {
-    auto final = this->get_canonical_state_from_ket(ket);
+    std::shared_ptr<const Type> final;
```
seweber: Use `final_state` instead of the reserved word `final`.

src/cpp/src/basis/BasisAtom.cpp
```diff
@@ -45,7 +46,13 @@ const std::string &BasisAtom<Scalar>::get_id_of_kets() const {
...
+    try {
```
seweber: These try-catch blocks look like a workaround to me, also elsewhere. Conceptually, we want a canonical state from a ket. Instead of getting it from `this`, the cleaner approach is probably to always construct it via `BasisAtomCreator<Scalar>().append_ket(ket).create(database)`. Alternatively, we could consider making get_matrix_elements work only for a basis, not for a ket. Then, it would be the task of the Python code to first convert a ket to a state (i.e., a basis object). I think I would prefer this solution.

src/cpp/src/basis/BasisPair.cpp
seweber, line 1: Could much of the logic from this file be moved into Python? I would be fine with get_matrix_elements only accepting a basis object on the C++ side if there are no significant performance downsides.

If helpful, we could also consider removing `get_matrix_elements` from the basis classes and making it a free function instead.

src/cpp/src/database/Database.cpp
```diff
@@ -942,6 +942,41 @@ Eigen::SparseMatrix<Scalar, Eigen::RowMajor> Database::get_matrix_elements_in_ca
...
+            "The initial and final bases must belong to the Database instance used for the "
+            "matrix element calculation.");
+    }
+    if (initial_basis->get_species() != final_basis->get_species()) {
+        throw std::invalid_argument(fmt::format(
+            "The initial and final bases must have the same species, but got '{}' and '{}'.",
+            initial_basis->get_species(), final_basis->get_species()));
+    }
+
+    if (type == OperatorType::IDENTITY) {
```
johannes-moegerle: make this analog to energy with database query

src/cpp/src/basis/BasisPair.cpp
```diff
@@ -114,28 +115,35 @@ template <typename Scalar>
 Eigen::VectorX<Scalar>
 BasisPair<Scalar>::get_matrix_elements(std::shared_ptr<const ket_t> ket, OperatorType type1,
                                        OperatorType type2, int q1, int q2) const {
-    // Construct a pair basis containing only the pair ket
-    auto final = this->get_canonical_state_from_ket(ket);
-    assert(final->get_number_of_states() == 1);
-
-    return this->get_matrix_elements(final, type1, type2, q1, q2).row(0);
+    try {
```
johannes-moegerle: actually we would like to remove this try catch.
We proceed as follows:
- first addd test cases/jupyter notebooks, to see when this is a bottlenck in performance
- then fix the performance issue

seweber (changes requested):

src/cpp/src/basis/BasisPair.cpp
seweber, line 1: In the file, there is also a variable named "final". I would search through the code to ensure "final" gets "final_state". To be consistent, we can also rename "initial" to "initial_state", "initial1" to "initial_state1", etc.

seweber (approved):

PR #426, opened 2026-06-15
C++ simplify get_matrix_elements

johannes-moegerle:
In C++ only allow get_matrix_elements between basis and basis.
Do handle the special cases like basis-ket overlap only in python

### DIFF
--- a/src/cpp/include/pairinteraction/basis/Basis.hpp
+++ b/src/cpp/include/pairinteraction/basis/Basis.hpp
@@ -18,7 +18,6 @@
 namespace pairinteraction {
 enum class Parity : int;
 enum class TransformationType : unsigned char;
-enum class OperatorType;
 
 /**
  * @class Basis
@@ -71,10 +70,6 @@ public:
     const Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &get_coefficients() const;
     Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &get_coefficients();
     void set_coefficients(const Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &values);
-    virtual Eigen::VectorX<scalar_t> get_matrix_elements(std::shared_ptr<const ket_t> ket,
-                                                         OperatorType type, int q = 0) const = 0;
-    Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> virtual get_matrix_elements(
-        std::shared_ptr<const Derived> other, OperatorType type, int q = 0) const = 0;
 
     class Iterator {
     public:
--- a/src/cpp/include/pairinteraction/basis/BasisAtom.hpp
+++ b/src/cpp/include/pairinteraction/basis/BasisAtom.hpp
@@ -13,6 +13,8 @@
 #include <vector>
 
 namespace pairinteraction {
+enum class OperatorType;
+
 class Database;
 
 class KetAtom;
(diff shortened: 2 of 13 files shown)


src/cpp/include/pairinteraction/basis/Basis.hpp
```diff
@@ -71,10 +70,6 @@ class Basis
     const Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &get_coefficients() const;
     Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &get_coefficients();
     void set_coefficients(const Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &values);
-    virtual Eigen::VectorX<scalar_t> get_matrix_elements(std::shared_ptr<const ket_t> ket,
-                                                         OperatorType type, int q = 0) const = 0;
-    Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> virtual get_matrix_elements(
```
seweber, line 76: I would keep this virtual method to ensure that every derived class has a get_matrix_elements.

src/cpp/include/pairinteraction/basis/BasisAtom.hpp
```diff
@@ -57,11 +59,8 @@ class BasisAtom : public Basis<BasisAtom<Scalar>>,
 
     int get_ket_index_from_id(size_t ket_id) const;
 
-    Eigen::VectorX<Scalar> get_matrix_elements(std::shared_ptr<const ket_t> ket, OperatorType type,
-                                               int q = 0) const override;
     Eigen::SparseMatrix<Scalar, Eigen::RowMajor>
-    get_matrix_elements(std::shared_ptr<const Type> other, OperatorType type,
-                        int q = 0) const override;
+    get_matrix_elements(std::shared_ptr<const Type> other, OperatorType type, int q) const;
```
seweber, line 63: I would keep `q=0`; otherwise, one has to set it for operators where one does not expect it. However, I do not have a strong opinion on it.

src/cpp/include/pairinteraction/basis/BasisPair.hpp
```diff
@@ -68,24 +68,9 @@ class BasisPair : public Basis<BasisPair<Scalar>>,
...
-                                               OperatorType type1, OperatorType type2, int q1 = 0,
-                                               int q2 = 0) const;
     Eigen::SparseMatrix<Scalar, Eigen::RowMajor>
     get_matrix_elements(std::shared_ptr<const Type> final_state, OperatorType type1,
-                        OperatorType type2, int q1 = 0, int q2 = 0) const;
-    Eigen::SparseMatrix<Scalar, Eigen::RowMajor>
-    get_matrix_elements(std::shared_ptr<const BasisAtom<Scalar>> final1,
-                        std::shared_ptr<const BasisAtom<Scalar>> final2, OperatorType type1,
-                        OperatorType type2, int q1 = 0, int q2 = 0) const;
+                        OperatorType type2, int q1, int q2) const;
```
seweber, line 73: Keep `q1=0` and `q2=0` as well if one keeps `q=0`.

src/cpp/tests/test_pair_potential.cpp
```diff
@@ -95,19 +95,25 @@ int main(int argc, char **argv) {
         kets.push_back(ss.str());
     }
 
+    auto basis_ket = pairinteraction::BasisAtomCreator<double>().add_ket(ket).create(database);
```
seweber, line 98: Looking at lines 98-101, I wonder whether we could improve the syntax. There is some friction with `add_ket` vs. `ket`, `create(database)` vs. `database()`.

seweber (changes requested):

seweber (approved):

PR #427, opened 2026-06-15
Improve database handling of column names

johannes-moegerle:
## TODOs in this PR
- add check in Database.cpp that every kets table has to include a id, f, and energy column and must not have a m column 

## Future related PRs done
- move get_label to python only function #436
- remove get_rotator and get_ket_for_different_quantum_number_m #433

## Future related PRs TODO
- remove C++ Parity enum completely
- Basis base class should not know about quantum numbers, only BasisAtom and BasisPair separately? similar to Ket classes in this PR
- get canonical_basis_id uniquely and deterministically (and wihtout collisions) (if we do so, we should b able to remove sql ket id)
- KetAtom remove non quantum number getters in python
- tolerance condition on if type of column is int or "f" or "m", or otherwise float


src/cpp/include/pairinteraction/ket/KetAtom.hpp
```diff
@@ -23,36 +22,29 @@ class KetAtom : public Ket {
...
 public:
-    KetAtom(Private /*unused*/, double energy, double f, double m, Parity p, std::string species,
-            int n, double nu, double nui_exp, double nui_std, double l_exp, double l_std,
-            double s_exp, double s_std, double j_exp, double j_std, double l_ryd_exp,
-            double l_ryd_std, double j_ryd_exp, double j_ryd_std, bool is_j_total_momentum,
-            bool is_calculated_with_mqdt, double underspecified_channel_contribution,
-            Database &database, size_t id_in_database);
+    /// The quantum numbers and additional meta information are stored in a single dictionary whose
+    /// keys are the column names of the database (e.g. "f", "m", "n", "nu", "exp_l", "std_l", ...).
+    /// Quantum numbers that come with a standard deviation are stored under the keys "exp_<name>"
```
johannes-moegerle: remove this comment?

src/cpp/src/basis/Basis.cpp
```diff
@@ -71,17 +71,27 @@ Basis<Derived>::Basis(ketvec_t &&kets)
     ket_to_ket_index.reserve(this->kets.size());
     size_t index = 0;
     for (const auto &ket : this->kets) {
-        state_index_to_quantum_number_f.push_back(ket->get_quantum_number_f());
-        state_index_to_quantum_number_m.push_back(ket->get_quantum_number_m());
-        state_index_to_parity.push_back(ket->get_parity());
+        real_t f = std::numeric_limits<real_t>::max();
```
johannes-moegerle, line 74: this pattern (and the changes in this file) is more a workaround, due to weird former coupling of BasisAtom and BasisPair assuming KetAtom and KetPair both! have m, f and parity.
Similar to how the Ket base class now only has energy and no m, f or parity, we maybe should also split this logic into BasisAtom and BasisPair separately?
Definitely needs a new PR

src/cpp/src/ket/KetAtom.cpp
```diff
@@ -18,73 +17,80 @@ namespace pairinteraction {
...
+    }
+    return 0;
+}
+
 std::string KetAtom::get_label() const {
+    double s = get_quantum_number("s");
+    double l = get_quantum_number("l");
+    double f = get_quantum_number("f");
+    double m = get_quantum_number("m");
+    bool is_calculated_with_mqdt = get_quantum_number("is_calculated_with_mqdt") != 0;
```
johannes-moegerle, line 47: how to handle get_label if we no longer guarantee this to be present?
Should we anyway move get_label to python only?

src/pairinteraction/enums.py
```diff
@@ -68,3 +68,21 @@ def get_python_parity(parity: _backend.Parity) -> Parity:
     if parity not in _ParityDict.values():
         raise ValueError(f"Unknown parity '{parity}', should be one of {list(_ParityDict.values())}")
     return next(k for k, v in _ParityDict.items() if v == parity)
+
+
+_ParityIntDict: dict[int, Parity] = {1: "even", -1: "odd", 2: "unknown"}
+
+
+def get_python_parity_from_int(parity: int) -> Parity:
```
johannes-moegerle: maybe overkill to define these functions? just directly handle it in KetAtom?

src/pairinteraction/basis/basis_atom.py
```diff
@@ -122,37 +122,25 @@ def __init__(  # noqa: C901, PLR0912, PLR0915
...
+            "l_ryd": l_ryd,
+            "j_ryd": j_ryd,
+            "f": f,
+            "m": m,
+        }
+        for name, value in quantum_numbers.items():
+            if value is not None:
+                creator.restrict_quantum_number(name, *value)
+                self._parameters_creator["qns"][name] = value
+        if parity is not None and parity != "unknown":
```
johannes-moegerle: the asymmetry of parity to the other quantum numbers is a bit unpleasant, but probably this is what we want to have here at the user facing interface

src/cpp/src/database/Database.cpp
```diff
@@ -636,50 +478,37 @@ Database::get_basis(const std::string &species, const AtomDescriptionByRanges &d
     // Ask the table for the extreme values of the quantum numbers
     {
         set_task_status("Validating atomic basis coverage...");
+
```
johannes-moegerle, line 517: actually, I am not sure, if this Validating atomic basis coverage is useful at all?

src/cpp/src/database/Database.cpp
```diff
@@ -38,13 +51,49 @@ std::string format_expectation_value_range(const std::string &value_column,
...
+double get_column_as_double(duckdb::Vector &vector, const duckdb::LogicalType &type, size_t row) {
+    switch (type.id()) {
+    case duckdb::LogicalTypeId::DOUBLE:
+        return duckdb::FlatVector::GetData<double>(vector)[row];
+    case duckdb::LogicalTypeId::BIGINT:
+        return static_cast<double>(duckdb::FlatVector::GetData<int64_t>(vector)[row]);
+    case duckdb::LogicalTypeId::BOOLEAN:
+        return duckdb::FlatVector::GetData<bool>(vector)[row] ? 1.0 : 0.0;
+    default:
+        throw std::runtime_error("Cannot read database column of type " + type.ToString() +
```
johannes-moegerle, line 67: should we throw if there is an unexpected column, or just skip that column?

src/cpp/bindings/ket/Ket.py.cpp
```diff
@@ -90,6 +61,7 @@ static void declare_ket_pair(nb::module_ &m, std::string const &type_name) {
     std::string pyclass_name = "KetPair" + type_name;
     nb::class_<KetPair<T>, Ket> pyclass(m, pyclass_name.c_str());
     pyclass.def("get_atomic_states", &KetPair<T>::get_atomic_states)
+        .def("get_quantum_number_m", &KetPair<T>::get_quantum_number_m)
```
seweber, line 64: Does KetPair really require this method? I would assume that the total m is relevant only for states.

src/cpp/include/pairinteraction/basis/BasisAtomCreator.hpp
```diff
@@ -38,35 +38,18 @@ class BasisAtomCreator {
...
-    BasisAtomCreator<Scalar> &restrict_quantum_number_nu(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_nui(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_l(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_s(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_j(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_l_ryd(real_t min, real_t max);
-    BasisAtomCreator<Scalar> &restrict_quantum_number_j_ryd(real_t min, real_t max);
+    // Set the quantum number range with the given logical name (e.g. "f", "m", "n", "l", ...).
+    // The parity is restricted via the name "parity" with a value of +1 (even) or -1 (odd).
+    BasisAtomCreator<Scalar> &restrict_quantum_number(const std::string &name, real_t min,
```
seweber, line 43: Can we remove the `Parity` enum if it is not used here, or would that have any drawbacks? I would assume that it would be sufficient to introduce a parity object on the Python side only.

src/cpp/include/pairinteraction/basis/BasisAtomCreator.hpp
```diff
@@ -38,35 +38,18 @@ class BasisAtomCreator {
...
     BasisAtomCreator<Scalar> &add_ket(const std::shared_ptr<const ket_t> &ket);
     std::shared_ptr<const BasisAtom<Scalar>> create(Database &database) const;
 
 private:
     std::optional<std::string> species;
-    Parity parity{Parity::UNKNOWN};
     Range<real_t> range_energy;
-    Range<real_t> range_quantum_number_f;
-    Range<real_t> range_quantum_number_m;
-    Range<int> range_quantum_number_n;
```
seweber, line 62: Can it lead to issues in comparisons/calculations if `n` is no longer stored as an int? I don't think so, but maybe we should cross-check. I have the same question for the parity.

src/cpp/include/pairinteraction/database/AtomDescriptionByRanges.hpp
```diff
@@ -3,27 +3,19 @@
...
-    Range<int> range_quantum_number_n;
-    Range<double> range_quantum_number_nu;
-    Range<double> range_quantum_number_nui;
-    Range<double> range_quantum_number_l;
-    Range<double> range_quantum_number_s;
-    Range<double> range_quantum_number_j;
-    Range<double> range_quantum_number_l_ryd;
-    Range<double> range_quantum_number_j_ryd;
+    // Quantum number ranges keyed by their logical name (e.g. "f", "m", "n", "l", ..., "parity").
+    std::unordered_map<std::string, Range<double>> quantum_number_ranges;
```
seweber, line 18: Just to be sure, is a range of parities handled correctly?

src/cpp/include/pairinteraction/ket/Ket.hpp
```diff
@@ -43,7 +35,7 @@ class Ket {
     }
 
 protected:
-    Ket(double energy, double f, double m, Parity p);
+    explicit Ket(double energy);
```
seweber, line 38: Why is `explicit` needed here?

src/cpp/include/pairinteraction/ket/KetAtom.hpp
```diff
@@ -23,36 +22,29 @@ class KetAtom : public Ket {
     struct Private {};
 
 public:
-    KetAtom(Private /*unused*/, double energy, double f, double m, Parity p, std::string species,
-            int n, double nu, double nui_exp, double nui_std, double l_exp, double l_std,
-            double s_exp, double s_std, double j_exp, double j_std, double l_ryd_exp,
-            double l_ryd_std, double j_ryd_exp, double j_ryd_std, bool is_j_total_momentum,
-            bool is_calculated_with_mqdt, double underspecified_channel_contribution,
-            Database &database, size_t id_in_database);
+    /// The quantum numbers and additional meta information are stored in a single dictionary whose
```
seweber: Wouldn't it be cleaner to have to dictionaries, one for exp and one for std?

src/cpp/include/pairinteraction/ket/KetPair.hpp
```diff
@@ -37,6 +37,8 @@ class KetPair : public Ket {
     std::string get_label() const override;
     std::shared_ptr<KetPair<Scalar>>
     get_ket_for_different_quantum_number_m(real_t new_quantum_number_m) const;
```
seweber, line 39: Does `get_ket_for_different_quantum_number_m` make it necessary that KetPair has a total m quantum number, or could we somehow rework the code so that it is sufficient for states having a total m quantum number?

src/cpp/src/basis/Basis.cpp
seweber: I completely agree. Especially KetPair should not have m, f, parity. Probably, basis could even avoid using m, f, parity of KetAtom. Could we make comments in the code to indicate the lines of the workaround so that we know what section of the code to address in a future PR?

src/cpp/src/basis/BasisAtomCreator.cpp
```diff
@@ -22,83 +21,14 @@ BasisAtomCreator<Scalar> &BasisAtomCreator<Scalar>::restrict_energy(real_t min,
...
-}
-
-template <typename Scalar>
-BasisAtomCreator<Scalar> &BasisAtomCreator<Scalar>::restrict_quantum_number_m(real_t min,
-                                                                              real_t max) {
-    if (2 * min != std::rint(2 * min) || 2 * max != std::rint(2 * max)) {
-        throw std::invalid_argument("Quantum number m must be an integer or half-integer.");
+BasisAtomCreator<Scalar> &
+BasisAtomCreator<Scalar>::restrict_quantum_number(const std::string &name, real_t min, real_t max) {
+    if ((name == "f" || name == "m") &&
```
seweber, line 26: Introduce a similar check for "parity".

src/cpp/src/ket/Ket.test.cpp
```diff
@@ -19,30 +18,31 @@ DOCTEST_TEST_CASE("constructing a class derived from ket") {
         struct Private {};
 
     public:
-        KetDerived(Private /*unused*/, double f, double m, Parity p) : Ket(0, f, m, p) {}
+        KetDerived(Private /*unused*/, double m) : Ket(0), quantum_number_m(m) {}
```
seweber: Maybe, remove the test case for constructing a KetDerived? It has diverged from how derived kets are constructed in the code, and maybe in the future, `get_ket_for_different_quantum_number_m` would also no longer be required.

src/cpp/src/ket/KetAtom.cpp
seweber: Yes, I would move it to Python only.

src/cpp/src/ket/KetAtomCreator.cpp
```diff
@@ -5,14 +5,17 @@
 
 #include "pairinteraction/database/AtomDescriptionByParameters.hpp"
 #include "pairinteraction/database/Database.hpp"
-#include "pairinteraction/enums/Parity.hpp"
 
 #include <cmath>
 
 namespace pairinteraction {
 KetAtomCreator::KetAtomCreator(std::string species, int n, double l, double j, double m)
```
seweber, line 12: I would remove this overload as it looks like convenience that does not belong in the C++ backend.

src/pairinteraction/basis/basis_atom.py
```diff
@@ -122,37 +122,25 @@ def __init__(  # noqa: C901, PLR0912, PLR0915
...
+            "f": f,
+            "m": m,
+        }
+        for name, value in quantum_numbers.items():
+            if value is not None:
+                creator.restrict_quantum_number(name, *value)
+                self._parameters_creator["qns"][name] = value
         if parity is not None:
-            creator.restrict_parity(get_cpp_parity(parity))
+            parity_int = parity_to_int(parity)
```
seweber, line 128: Could there be a more elegant solution? My guess for the best approach would be to remove the parity enum from C++ and convert the Python parity object to an int via a nanobind binding.

src/pairinteraction/ket/ket_atom.py
```diff
@@ -155,87 +164,87 @@ def species(self) -> str:
...
     def is_j_total_momentum(self) -> bool:
         """Whether j is the total momentum quantum number, otherwise f is the total momentum quantum number."""
-        return self._cpp.is_j_total_momentum()
+        return bool(self._cpp.get_quantum_number("is_j_total_momentum"))
 
     @property
     def is_calculated_with_mqdt(self) -> bool:
         """Whether the state was calculated with multi-channel quantum defect theory."""
-        return self._cpp.is_calculated_with_mqdt()
+        return bool(self._cpp.get_quantum_number("is_calculated_with_mqdt"))
```
seweber, line 242: I would already now remove everything that is not a quantum number.

src/pairinteraction/ket/ket_pair.py
```diff
@@ -46,6 +46,11 @@ def __init__(self) -> None:
         """Creating a KetPair object directly is not possible."""  # noqa: D401
         raise NotImplementedError("KetPair objects cannot be created directly.")
 
+    @property
```
seweber, line 49: I would already now remove the total m of KetPair at least from the binding.

src/pairinteraction/enums.py
```diff
@@ -34,12 +34,9 @@
     "arbitrary": _backend.OperatorType.ARBITRARY,
 }
 
-Parity = Literal["even", "odd", "unknown"]
-_ParityDict: dict[Parity, _backend.Parity] = {
-    "even": _backend.Parity.EVEN,
-    "odd": _backend.Parity.ODD,
-    "unknown": _backend.Parity.UNKNOWN,
-}
+Parity = Literal["even", "odd"]
```
seweber, line 37: Probably, the handling of the parity can be simplified significantly. See my comment above.

src/cpp/src/database/Database.cpp
```diff
@@ -38,13 +43,49 @@ std::string format_expectation_value_range(const std::string &value_column,
...
+size_t get_column_index(const std::vector<std::string> &names, const std::string &name) {
+    auto it = std::find(names.begin(), names.end(), name);
+    if (it == names.end()) {
+        throw std::runtime_error("Missing database column '" + name + "'.");
     }
+    return static_cast<size_t>(std::distance(names.begin(), it));
+}
+
+// Read a single value of a duckdb result column as a double, regardless of its logical type.
+double get_column_as_double(duckdb::Vector &vector, const duckdb::LogicalType &type, size_t row) {
```
seweber: get column or get entry?

src/cpp/src/database/Database.cpp
```diff
@@ -38,13 +43,49 @@ std::string format_expectation_value_range(const std::string &value_column,
...
+                                 " as a quantum number.");
+    }
+}
+
+// Build the quantum-number map of a KetAtom from a query result row, treating every column except
+// the given ones as a quantum number keyed by its database column name.
+std::unordered_map<std::string, double>
+make_quantum_numbers(duckdb::DataChunk &chunk, const std::vector<duckdb::LogicalType> &types,
+                     const std::vector<std::string> &names,
+                     const std::unordered_set<std::string> &excluded_columns, size_t row) {
```
seweber: Or should we hardcode "energy" etc. instead of using excluded_columns? I would assume that the columns to exclude are fundamentally fixed.

src/cpp/src/database/Database.cpp
seweber: If I remember correctly, the idea was to provide the user with feedback if no states are stored in the table for the specified parameter range. Otherwise, users might not realize that they have only obtained a part of their specified basis.

src/cpp/include/pairinteraction/ket/KetAtom.hpp
seweber: Yes

src/pairinteraction/basis/basis_atom.py
```diff
@@ -122,37 +122,25 @@ def __init__(  # noqa: C901, PLR0912, PLR0915
...
+            "n": n,
+            "nu": nu,
+            "nui": nui,
+            "s": s,
+            "l": l,
+            "j": j,
+            "l_ryd": l_ryd,
+            "j_ryd": j_ryd,
+            "f": f,
+            "m": m,
```
johannes-moegerle, line 119: add parity to this dictionary

src/cpp/src/database/Database.cpp
seweber: Throw is fine with me.

src/cpp/src/basis/Basis.cpp
```diff
@@ -71,17 +71,27 @@ Basis<Derived>::Basis(ketvec_t &&kets)
...
+        real_t m = std::numeric_limits<real_t>::max();
+        Parity p = Parity::UNKNOWN;
+        if constexpr (requires { ket->get_quantum_number(std::string{}); }) {
```
johannes-moegerle, line 79: add comment with workaround and todo (also other places here)

src/cpp/src/ket/Ket.test.cpp
johannes-moegerle, line 1: remove this file

src/cpp/include/pairinteraction/ket/KetAtom.hpp
johannes-moegerle: https://github.com/pairinteraction/pairinteraction/pull/427/changes/44c4427f2bd5465998bde65057bc76e80f92da1c

fine like that?

seweber (approved):

PR #428, opened 2026-06-16
Allow merge basis

johannes-moegerle:


src/cpp/src/basis/BasisPair.cpp
```diff
@@ -49,6 +53,75 @@ int BasisPair<Scalar>::get_ket_index_from_tuple(size_t state_index1, size_t stat
...
+    const auto atomic_bases_are_compatible = [](const auto &basis_a, const auto &basis_b) {
+        return basis_a->has_same_canonical_basis(basis_b) &&
+            basis_a->get_number_of_kets() == basis_b->get_number_of_kets() &&
+            basis_a->get_number_of_states() == basis_b->get_number_of_states() &&
+            basis_a->get_coefficients().isApprox(basis_b->get_coefficients());
+    };
+    if (!atomic_bases_are_compatible(basis1, other->basis1) ||
+        !atomic_bases_are_compatible(basis2, other->basis2)) {
+        throw std::invalid_argument(
+            "Cannot merge pair bases with different ordered atomic states or transformations.");
```
seweber: I am not sure how well this error message can be understood by a user.

tests/test_state_atom.py
```diff
@@ -87,6 +87,57 @@ def test_get_matrix_element(pi_module: PairinteractionModule, basis: BasisAtom)
...
+    assert pytest.approx(combined.get_overlap(ket2)) == 0.5  # NOSONAR
+
+
+def test_add_states_with_different_bases_merges(pi_module: PairinteractionModule) -> None:
+    """Adding states with different bases merges the bases and re-expresses the coefficients."""
+    ket1 = pi_module.KetAtom("Rb", n=60, l=0, j=0.5, m=0.5)
+    ket2 = pi_module.KetAtom("Rb", n=60, l=1, j=1.5, m=0.5)
+
+    # Both states start from a minimal single-ket basis, so their bases differ.
+    state1 = pi_module.StateAtom(ket1)
```
seweber, line 119: I wonder whether the following syntax could be more user-friendly: `state1 = pi_module.KetAtom("Rb", n=60, l=0, j=0.5, m=0.5).to_state()`.

In addition, I have the feeling that we have a mixture between situations where ket gets automatically converted to state and situations where the user has to do it. I see three possible ways to go forward: i) Maybe we could let, e.g., the function that calculates the overlap only take state objects? Then, a user would have to pass ket.to_state(). ii) Alternatively, we could do auto conversion for, e.g., `ket1 + ket2`. iii) Or, maybe, ket does not need to be user-facing at all?

src/pairinteraction/basis/basis_base.py
```diff
@@ -139,6 +139,30 @@ def canonicalized(self: Self) -> Self:
...
+        For BasisPair they also must share the same underlying atomic basis.
+        The returned basis is in canonical form, i.e. has identity coefficients.
+        Consequently, coefficients of transformed or symmetrized input bases are not retained.
+        """
+        if type(self) is not type(other):
+            raise TypeError(
+                f"Can only merge {type(self).__name__} with the same basis type, but got {type(other).__name__}."
+            )
+        if not self.is_canonical or not other.is_canonical:
+            logger.warning(
```
seweber: Shouldn't merge be allowed only on canonical bases, and shouldn't this be already ensured in C++? Especially since a user can call basis.canonicalized() to easily obtain a canonical basis, I do not see a downside to making this the user's responsibility.

src/pairinteraction/state/state_atom.py
```diff
@@ -75,11 +88,20 @@ def __add__(self, other: Self) -> Self:
...
 
-        coeffs = self._cpp.get_coefficients() + other._cpp.get_coefficients()
-        new_cpp = self._cpp.copy()
+        # A shared canonical basis means both states are expressed with the exact same kets in the same order
+        if self._cpp.has_same_canonical_basis(other._cpp):
+            coeffs = self._cpp.get_coefficients() + other._cpp.get_coefficients()
+            new_cpp = self._cpp.copy()
+        else:
+            # merge the basis and re-express the coefficients in the merged basis
+            merged_cpp = self._cpp.merge(other._cpp)
```
seweber: Shouldn't canonicalized() be called on the states before merging them?

src/pairinteraction/state/state_atom.py
```diff
@@ -75,11 +88,20 @@ def __add__(self, other: Self) -> Self:
...
+            cpp_op = get_cpp_operator_type("identity")
```
seweber: Do we have `get_cpp_[...](string)` calls also for converting other python literals, or is  `get_cpp_operator_type("identity")` special?

src/pairinteraction/state/state_atom.py
```diff
@@ -75,11 +88,20 @@ def __add__(self, other: Self) -> Self:
...
-        if not all(ket_self == ket_other for ket_self, ket_other in zip(self.kets, other.kets, strict=True)):
-            raise ValueError("Cannot add states where the basis does not consist of exactly the same kets.")
 
-        coeffs = self._cpp.get_coefficients() + other._cpp.get_coefficients()
-        new_cpp = self._cpp.copy()
+        # A shared canonical basis means both states are expressed with the exact same kets in the same order
+        if self._cpp.has_same_canonical_basis(other._cpp):
+            coeffs = self._cpp.get_coefficients() + other._cpp.get_coefficients()
+            new_cpp = self._cpp.copy()
+        else:
```
seweber: Couldn't there be a case where the states neither have the same canonical bases nor independent bases? If this is the case, I would expect the "else" branch to handle this situation. If this else branch does it, it should also be able to handle the same canonical bases correctly, and the branching is not needed (except maybe for performance).

src/cpp/src/basis/BasisAtom.cpp
```diff
@@ -42,6 +45,34 @@ const std::string &BasisAtom<Scalar>::get_canonical_basis_id() const {
     return canonical_basis_id;
 }
 
+template <typename Scalar>
+bool BasisAtom<Scalar>::has_same_canonical_basis(std::shared_ptr<const Type> other) const {
```
seweber: Is this method really needed (see my comment in state_atom.py where it is called)?

src/cpp/src/basis/BasisPair.cpp
```diff
@@ -49,6 +53,75 @@ int BasisPair<Scalar>::get_ket_index_from_tuple(size_t state_index1, size_t stat
     return state_indices_to_ket_index.at({state_index1, state_index2});
 }
 
+template <typename Scalar>
+std::shared_ptr<const typename BasisPair<Scalar>::Type>
+BasisPair<Scalar>::merge(std::shared_ptr<const Type> other) const {
```
seweber, line 58: This implementation seems to perform the merge differently from BasisAtom::merge. Why can't we merge here canonicalized bases as well?

src/cpp/src/basis/BasisPair.cpp
```diff
@@ -49,6 +53,75 @@ int BasisPair<Scalar>::get_ket_index_from_tuple(size_t state_index1, size_t stat
...
+    const auto atomic_bases_are_compatible = [](const auto &basis_a, const auto &basis_b) {
+        return basis_a->has_same_canonical_basis(basis_b) &&
+            basis_a->get_number_of_kets() == basis_b->get_number_of_kets() &&
+            basis_a->get_number_of_states() == basis_b->get_number_of_states() &&
+            basis_a->get_coefficients().isApprox(basis_b->get_coefficients());
```
seweber, line 63: Requiring two bases to have the same canonical bases and the same coefficients looks strange. Aren't these the same bases then?

seweber (changes requested):

src/cpp/src/basis/BasisPair.cpp
johannes-moegerle: `"Cannot merge pair bases with different single atom bases objects."` better?

tests/test_state_atom.py
johannes-moegerle: In principle I also like the `ket.to_state()` syntax, I am just not sure if we then completely should drop the StateAtom init, which I dont like, or directly let it take a list of kets and coefficients?

Do you have an example, where a users has to convert the ket to a state by its own?
I think my perspective was always allow go get a ket instead of a state, but in addition allow the user also to create a "more complicated state" which then can be passed as state to e.g. basis.get_overlaps, etc
i) I think allowing to pass kets and then basically run ket.to_state() as convenience for the user is fine on python level.
ii) `ket1 + ket2` is currently not allowed, I am fine with it, however it makes the ket to state conversion more implicit
iii) I disagree, I think we should keep the underlying structure of kets and states, which are linear combinations of kets, transparent to the user, which would be more hidden if you somehow try to make KetAtom not userfacing

src/pairinteraction/basis/basis_base.py
johannes-moegerle: true

src/pairinteraction/state/state_atom.py
johannes-moegerle: yes

src/pairinteraction/state/state_atom.py
johannes-moegerle: yes, also exists at other places for other operators

src/cpp/src/basis/BasisPair.cpp
johannes-moegerle: yes, to be able to merge two BasisPair objects, their SystemAtom (and BasisAtoms) have to be the same.
Maybe we really could require them to be the same object instead of checking that they have the same kets and same coefficients, which however is a bit more general ...

src/pairinteraction/state/state_atom.py
johannes-moegerle: yes, in principle the else branch can already handle all cases.
Probably the performance here is not necessary, so I will remove the if part and just keep the general else part

src/cpp/src/basis/BasisPair.cpp
johannes-moegerle: not quite sure what you mean, the BasisPair constructions is different from BasisAtom construction, which is why also the merge here is different

src/cpp/src/basis/BasisPair.cpp
johannes-moegerle: "Cannot merge two pair bases which where constructed from different pairs of SystemAtom objects."

tests/test_state_atom.py
johannes-moegerle: allow `ket1 + ket2` and `ket.to_state()`, deprecate the old `StateAtom` constructor, maybe make the new one as `(kets, coefficients (basis?))`
(extra PR)

seweber (approved):

src/cpp/src/basis/BasisPair.cpp
johannes-moegerle: new PR to allow identical basis check

PR #431, opened 2026-06-18
Improve BasisPairCreator with symmetries

johannes-moegerle:


src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -70,6 +69,13 @@ std::shared_ptr<const BasisPair<Scalar>> BasisPairCreator<Scalar>::create() cons
     const bool has_symmetry_restriction =
         parity_under_inversion != Parity::UNKNOWN || parity_under_permutation != Parity::UNKNOWN;
 
+    // This ensures, that a one-atom state can be identified across both atoms by its state index
+    if (has_symmetry_restriction && &systems_atom[0].get() != &systems_atom[1].get()) {
```
seweber, line 73: Just to be sure, does it make sense to compare the systems by their pointers? I think so, but I am not completely certain.

src/cpp/src/basis/BasisPairCreator.cpp
seweber, line 1: I am not sure I understand why `record_symmetrized_contribution` was introduced (it does not change anything compared to the previous implementation, does it?). However, I am open to it.

seweber (approved):

PR #432, opened 2026-06-19
remove ket-state mapping in basis

johannes-moegerle:


johannes-moegerle:
replaced by #436 and #437

PR #435, opened 2026-06-26
Suppress qt.gui.icc.warning

seweber:


src/pairinteraction_gui/__init__.py
```diff
@@ -18,6 +19,10 @@ def main(*, enable_theme_hot_reload: bool = False) -> int:
         int: Application exit code
 
     """
+    rules = os.environ.get("QT_LOGGING_RULES", "")
+    if "qt.gui.icc.warning=false" not in rules.splitlines():
```
johannes-moegerle: why this if clause?
Either you want `"qt.gui.icc.warning" not in rules.splitlines():` or, if you always want to override it just always use `QLoggingCategory.setFilterRules(f"{rules}\nqt.gui.icc.warning=false".strip())`?

johannes-moegerle (approved):
squash and rebase and then fine to merge

PR #436, opened 2026-07-01
move ket get label to python

### LINKED ISSUE #402: ket naming convention inconsistency
Pure triplet or singlet states seem to follow a different naming convention than mixed states. I think there is potential for confusion here, because this could be interpreted as states being taken from different databases (compare the example below, where "Yb171_triplet" and "Yb171" are actually from the same database).

Example

`ket1 = pi.KetAtom("Yb171_mqdt", nu=55.5, l=0, f=1.5, m=1.5)`
`ket2 = pi.KetAtom("Yb171_mqdt", nu=55.1, l=1, f=2.5, m=2.5)`
`print(ket1)`
`print(ket2)`

`|Yb171_triplet:60,S_3/2,3/2⟩`
`|Yb171:S=1.0,nu=55.1,L=1.0,F=5/2,5/2⟩`

johannes-moegerle:
- in addition this fixes #402

### DIFF
--- a/src/cpp/bindings/ket/Ket.py.cpp
+++ b/src/cpp/bindings/ket/Ket.py.cpp
@@ -17,7 +17,6 @@
 #include <nanobind/stl/shared_ptr.h>
 #include <nanobind/stl/string.h>
 #include <nanobind/stl/vector.h>
-#include <sstream>
 
 namespace nb = nanobind;
 using namespace pairinteraction;
@@ -25,13 +24,7 @@ using namespace pairinteraction;
 static void declare_ket(nb::module_ &m) {
     std::string pyclass_name = "Ket";
     nb::class_<Ket> pyclass(m, pyclass_name.c_str());
-    pyclass.def("get_energy", &Ket::get_energy)
-        .def("get_label", &Ket::get_label)
-        .def("__str__", [](Ket const &self) {
-            std::stringstream ss;
-            ss << self;
-            return ss.str();
-        });
+    pyclass.def("get_energy", &Ket::get_energy);
 }
 
 static void declare_ket_atom(nb::module_ &m) {
--- a/src/cpp/bindings/ket/Ket.py.cpp
+++ b/src/cpp/bindings/ket/Ket.py.cpp
@@ -8,6 +8,7 @@
 #include "pairinteraction/ket/Ket.hpp"
 #include "pairinteraction/ket/KetAtom.hpp"
 #include "pairinteraction/ket/KetAtomCreator.hpp"
+#include "pairinteraction/ket/KetNotUniqueError.hpp"
 #include "pairinteraction/ket/KetPair.hpp"
 #include "pairinteraction/utils/traits.hpp"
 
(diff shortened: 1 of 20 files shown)


src/cpp/bindings/ket/Ket.py.cpp
```diff
@@ -66,10 +60,27 @@ static void declare_ket_pair(nb::module_ &m, std::string const &type_name) {
         .def("__hash__", [](const KetPair<T> &self) { return typename KetPair<T>::hash{}(self); });
 }
 
+static void declare_ket_not_unique_error(nb::module_ &m) {
+    static nb::object exc_type = nb::exception<KetNotUniqueError>(m, "KetNotUniqueError");
+    nb::register_exception_translator(
+        [](const std::exception_ptr &p, void *payload) {
+            try {
+                std::rethrow_exception(p);
```
seweber, line 68: Do I understand it correctly that `rethrow_exception` followed by `catch` is basically a trick to figure out which type of exception was thrown? To me, it looks a bit hacky, and I wonder whether this is the default pattern for making nanobind aware of custom C++ exceptions. It might well be, but I wanted to double-check.

src/cpp/src/database/Database.cpp
```diff
@@ -380,9 +381,10 @@ std::shared_ptr<const KetAtom> Database::get_ket(const std::string &species,
             get_entry_as_double(chunk->data[order_val_column], types[order_val_column], 1);
 
         if (order_val_1 - order_val_0 <= order_val_0) {
-            throw std::invalid_argument(
-                fmt::format("The ket is not uniquely specified. Possible kets are:\n{}\n{}",
-                            fmt::streamed(make_ket(0)), fmt::streamed(make_ket(1))));
+            std::vector<std::shared_ptr<const KetAtom>> candidates;
+            candidates.push_back(std::make_shared<const KetAtom>(make_ket(0)));
+            candidates.push_back(std::make_shared<const KetAtom>(make_ket(1)));
+            throw KetNotUniqueError(std::move(candidates));
```
seweber: Optionally, this could be simplified to the following code if I am not mistaken
```c++
throw KetNotUniqueError({
    std::make_shared<const KetAtom>(make_ket(0)),
    std::make_shared<const KetAtom>(make_ket(1)),
});
```

src/cpp/tests/test_pair_potential.cpp
```diff
@@ -81,20 +81,12 @@ int main(int argc, char **argv) {
     }
 
     // Extract results
```
seweber, line 83: Could we instead perform an ID-based check to ensure the kets match those in the reference data (also in test_starkmap.cpp)? I think this can be useful, as such a check can help investigate issues, e.g., emerging from a change in the order of the kets.

seweber (approved):

src/cpp/bindings/ket/Ket.py.cpp
johannes-moegerle: yes kind of, I think the main "problem" why this pattern is needed is to bind the possible kets as attrbitue to the error, so I can reuse them in python to throw an error with properly formatted labels of potential kets. 
see also https://nanobind.readthedocs.io/en/latest/exceptions.html#handling-custom-exceptions

PR #437, opened 2026-07-01
Remove C++ get corresponding methods

johannes-moegerle:


src/cpp/include/pairinteraction/enums/TransformationType.hpp
```diff
@@ -9,7 +9,7 @@
 namespace pairinteraction {
 enum class TransformationType : unsigned char {
     IDENTITY = 0,
-    SORT_BY_KET = 1 << 0,
+    CANONICAL_ORDER = 1 << 0,
```
seweber, line 12: I would remove `CANONICAL_ORDER` here as it is no sorting, see `is_sorting` and how `canonicalized()` is implemented.  Instead, I would store `TransformationType` in an std::optional type within `coefficients`, or not store it at all, or rename `CANONICAL_ORDER` to `NONE`. I am not sure what the best approach is. If not storing at all is viable, I would go for it.

src/pairinteraction/state/state_base.py
```diff
@@ -136,3 +136,30 @@ def get_overlap(self, other: Any) -> Any: ...
 
     @abstractmethod
     def get_matrix_element(self, other: Any, *args: Any, **kwargs: Any) -> Any: ...
+
+
+def _get_corresponding_index(overlaps: NDArray, subject: str, target: str) -> int:
```
seweber: I find the method a bit confusing. The argument `target` looks as if state and ket are handled differently. Maybe rename it to something like error_string and pass "{target} for {subject}" to it?

I would also consider renaming "_get_corresponding_index" to somthing like "_get_indices_with_largest_overlaps".

src/pairinteraction/state/state_base.py
```diff
@@ -136,3 +136,30 @@ def get_overlap(self, other: Any) -> Any: ...
...
+    if len(overlaps) == 0:
+        raise ValueError(f"Cannot find corresponding {target}: No overlaps (probably due to empty basis).")
+    if len(overlaps) == 1:
+        largest_id = 0
+        largest_overlap = overlaps[0]
+        second_largest_overlap = 0
+    else:
+        ids = np.argpartition(overlaps, -2)[-2:]
```
seweber, line 152: I am not sure whether I understand it. Naively, I would have assumed using np.argmax along some axis.

src/pairinteraction/state/state_pair.py
```diff
@@ -159,3 +155,6 @@ def get_matrix_element(self, other: Any, *args: Any, **kwargs: Any) -> Any:
 class StatePairReal(StatePair):
     _cpp: _backend.BasisPairReal  # type: ignore [assignment]
     _ket_class = KetPairReal
+
+
+StatePairLike = TypeAliasType("StatePairLike", StatePair | tuple[StateAtom, StateAtom] | Sequence[StateAtom])
```
seweber: Shouldn't the definitions of KetPairLike and StatePairLike be more analogous to each other? I find it a bit confusing that there is a KetAtomTuple but no StateAtomTuple.

seweber (changes requested):

src/pairinteraction/state/state_base.py
johannes-moegerle: add a comment

seweber (approved):

PR #444, opened 2026-07-22
Improve Basis

johannes-moegerle:


src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -27,7 +26,16 @@ BasisPairCreator<Scalar> &BasisPairCreator<Scalar>::add(const SystemAtom<Scalar>
     if (!system_atom.is_diagonal()) {
         throw std::invalid_argument("The system must be diagonalized before it can be added.");
     }
-    systems_atom.push_back(system_atom);
+    // The eigenstates must be sorted by energy. This is ensured by System::diagonalize and
+    // required for the binary search of the energetically allowed range in create().
+    auto eigenenergies = system_atom.get_eigenenergies();
+    if (!std::is_sorted(eigenenergies.data(), eigenenergies.data() + eigenenergies.size())) {
```
seweber: I would consider implementing this evaluation in a `system_atom.is_sorted_by_energy()` method, analogous to `system_atom.is_diagonal()` (and then calling this method internally in System.cpp as well).

src/cpp/src/system/System.cpp
```diff
@@ -220,6 +220,10 @@ System<Derived> &System<Derived>::diagonalize(const DiagonalizerInterface<scalar
     }
 
     if (this->is_diagonal()) {
```
seweber, line 223: From my point of view, `is_diagonal` sounds too similar to `is_diagonalized`. As diagonalization now implies sorting by energy, this might be confusing because `is_diagonal` doesn't check the sorting. Thus, to avoid confusion, I would rename `is_diagonal` to something like `is_diagonal_hamiltonian`. Maybe, one could also improve the situation with some more structural changes. However, so far, I do not see an obvious structural change that would help.

src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -61,19 +69,26 @@ template <typename Scalar>
...
+    // Physically, symmetrization is only defined for two identical atoms, which is enforced by this
+    // check. This automatically also ensures that a one-atom state can be identified across both
+    // atoms by its state index.
+    if (has_symmetry_restriction) {
+        const bool have_same_eigenenergies =
+            eigenenergies_atom[0].size() == eigenenergies_atom[1].size() &&
+            (eigenenergies_atom[0] - eigenenergies_atom[1]).isZero(numerical_precision);
+        if (bases_atom[0] != bases_atom[1] || !have_same_eigenenergies) {
+            throw std::invalid_argument(
+                "Parity restrictions require the same SystemAtom, i.e. same basis and same "
```
seweber: Hmm, I don't know whether I like comparing the bases and energies instead of the system pointers. Isn't comparing the system pointers closer to the physical picture of having two identical atoms? One also avoids some unnecessary copies and O(n) operations if one performs a comparison based on the systems.

seweber (changes requested):

src/cpp/src/basis/BasisPairCreator.cpp
johannes-moegerle: `system_atom.is_diagonal_and_sorted_by_energy`
und diagonalize add sort_by_energy argument (default True) in C++

src/cpp/src/basis/BasisPairCreator.cpp
johannes-moegerle: leave it here with system atoms, make a new PR to add energy_of_states to basis objects (together m_of_states, parity_of_states clean up), and after that make it possible/the intended behaviour to construct a basispair from two basisatoms which both have energy_of_states

seweber (approved):

PR #448, opened 2026-07-29
Cleanup transformation type

johannes-moegerle:
# Remove the recorded transformation type

`Transformation` and `Sorting` each carried a `transformation_type` — a list of
`TransformationType` values recording how a basis or system had been transformed. This was
redundant bookkeeping that could get out of sync with the actual coefficients, and it was only ever
read by two checks. Both checks are now derived from the data itself, which lets the whole mechanism
go away. Removing it also uncovered two bugs in `get_sorter`, fixed here.

## What changed

**`Basis::is_canonical` inspects the coefficient matrix.** "Canonical" means the *i*-th state is the
*i*-th ket, i.e. the coefficient matrix is the identity. That is now checked directly instead of
comparing against a recorded `CANONICAL_ORDER` label.

**`Basis::perform_blocks_checks` inspects the state labels.** Instead of asking whether the recorded
transformation type mentions the requested labels, it checks the property that the block
decomposition actually needs: states sharing the same labels must be contiguous. If they were
scattered over several blocks, the couplings between them would be lost when diagonalizing block by
block. The check also verifies up front that the states are labeled by the requested quantum numbers
at all.

**The recorded transformation type is removed**, and with it `get_transformation()`, which returned
the very same coefficients as `get_coefficients()`, only wrapped.

**`CANONICAL_ORDER` and `ARBITRARY` are removed from the enum.** They only existed as recorded
transformation types. With every remaining value being a sorting label, `utils::is_sorting` is always
true and `perform_sorter_checks` can no longer reject anything, so both are dropped too.

**`Transformation` and `Sorting` are removed.** Without the recorded type they only wrapped a
matrix. `Eigen::SparseMatrix` and `Eigen::PermutationMatrix` are used directly — they are distinct
types, so the two `transform`/`transformed` overloads stay distinguishable.

**`SorterBuilderInterface` is no longer a template.** It was templated on the scalar type only
because `get_transformation()` returned a `Transformation<Scalar>`. Nothing that is left depends on
the scalar type, so the template parameter, the `static_assert`, the `real_t` alias and the two
explicit instantiations are gone, and the two bound python classes collapse into one.

**Renames to match what is left:** `TransformationType` → `SorterType`,
`TransformationBuilderInterface` → `SorterBuilderInterface`. The interface no longer builds
transformations; it declares `get_sorter` and `get_indices_of_blocks`. The enum values lose their
now redundant prefix, e.g. `SORT_BY_ENERGY` → `SorterType::ENERGY`.

## Bug fix: sorting by several labels

`get_sorter` is documented to treat the first label as the primary sorting criterion, with the
following labels only breaking ties. It applies the labels as successive stable sorts, so the sort
applied *last* determines the primary criterion. The label groups around the energy were applied in
the order they were given rather than in reverse, so the precedence was inverted:
`get_sorter({QUANTUM_NUMBER_M, ENERGY})` sorted primarily by the energy instead of by *m*.

A repeated label also used to terminate the process. `get_sorter({ENERGY, ENERGY})` split the labels
at the first energy, leaving a second energy label in the group handed to the basis, which reached a
`std::abort()` in the branch for labels the basis cannot handle. Labels are now deduplicated first;
a repeated label cannot influence the order anyway.

## Bug fix: blocks of a Hamiltonian by energy

`System::get_indices_of_blocks` accepted `SORT_BY_ENERGY` and built blocks from runs of equal
diagonal entries, comparing floating-point energies with `!=`. Such blocks are only meaningful for an
already diagonal Hamiltonian, which needs no block-by-block diagonalization in the first place.
Requesting energy blocks is now rejected with a clearer message; sorting a system by the energy
remains supported.

## Tests

`SystemAtom.test.cpp` gains coverage for block-diagonalization by *m* under an electric field, for
the rejection of energy blocks while sorting by energy keeps working, for repeated labels, and for
the precedence of several labels — including that the energy is only a tie breaker when it is not
the first label.

`BasisAtomCreator.test.cpp` gains coverage for `is_canonical` (identity coefficients, explicitly
stored zeros, a non-trivial superposition, a sorted basis) and for the preconditions of
`get_indices_of_blocks`: an unsorted basis is rejected, a basis sorted by parity and *m* yields
blocks for both together and for the primary criterion alone but not for *m* alone, and an unlabeled
basis is rejected.

## API impact

C++ only — no changes outside `src/cpp`. No python-level code in the repository used any of the
removed names, but the private `_backend` module changes:

- `TransformationType` is now `SorterType`, and its values lose the `SORT_BY_` prefix.
- `Transformation{Real,Complex}`, `Sorting`, `TransformationBuilderInterface{Real,Complex}` and
  `get_transformation` are gone. `transform`/`transformed` take an Eigen sparse matrix or a
  `PermutationMatrix` directly.
- `System.get_indices_of_blocks` raises for the energy label instead of returning degeneracy blocks.
- `PermutationMatrix` is no longer constructible; it is only ever obtained from `get_sorter`. The
  `IndicesOfBlocksCreator` class and `Basis.get_indices_of_blocks_without_checks` are unbound, the
  latter having been uncallable without a constructible creator.

# Future PRs
- clean up names of `get_sorter` and `get_sorter_without_checks`, for this maybe it is beneficial to add `energy_of_states` to the basis objects?


src/cpp/bindings/system/System.py.cpp
```diff
@@ -33,27 +33,23 @@ static void declare_system(nb::module_ &m, const std::string &type_name) {
...
-            },
+            [](S &self, const Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> &transformation)
+                -> T & { return static_cast<T &>(self.transform(transformation)); },
             nb::call_guard<nb::gil_scoped_release>())
         .def(
             "transform",
-            [](S &self, const Sorting &sorting) -> T & {
-                return static_cast<T &>(self.transform(sorting));
-            },
+            [](S &self, const Eigen::PermutationMatrix<Eigen::Dynamic, Eigen::Dynamic> &sorter)
```
seweber, line 51: What possibilities exist to obtain a PermutationMatrix in Python? Is `get_sorter` the only way, or could a user build their own permutation matrix with a low-level interface (not sure whether this would necessarily be desirable)?

src/cpp/include/pairinteraction/system/System.hpp
```diff
@@ -57,7 +58,7 @@ class System
     mutable Eigen::SparseMatrix<scalar_t, Eigen::RowMajor> matrix;
     mutable bool hamiltonian_requires_construction{true};
     mutable bool hamiltonian_is_diagonal{false};
-    mutable std::vector<TransformationType> blockdiagonalizing_labels;
+    mutable std::vector<SorterType> blockdiagonalizing_labels;
```
seweber, line 61: Energy is a SorterType but not a blockdiagonalizing label.

src/cpp/src/basis/Basis.cpp
```diff
@@ -6,62 +6,84 @@
...
-    // Check if the labels are valid sorting labels
-    for (const auto &label : labels) {
-        if (label == TransformationType::CANONICAL_ORDER) {
-            throw std::invalid_argument("Sorting by canonical order is not supported.");
-        }
-        if (!utils::is_sorting(label)) {
-            throw std::invalid_argument("One of the labels is not a valid sorting label.");
-        }
+void Basis<Derived>::perform_blocks_checks(const std::set<SorterType> &unique_labels) const {
+    // Blocks of states of equal energy would only be blocks of an already diagonal Hamiltonian,
```
seweber: Would allowing for ENERGY really be a problem?

src/cpp/src/basis/Basis.cpp
```diff
@@ -242,159 +263,148 @@ typename Basis<Derived>::Iterator &Basis<Derived>::Iterator::operator++() {
...
+    const std::vector<SorterType> &labels,
+    Eigen::PermutationMatrix<Eigen::Dynamic, Eigen::Dynamic> &sorter) const {
     constexpr real_t numerical_precision = 100 * std::numeric_limits<real_t>::epsilon();
 
-    int *perm_begin = transformation.matrix.indices().data();
-    int *perm_end = perm_begin + coefficients.matrix.cols();
+    // Throw a meaningful error if sorting by the energy is requested as this might be a common
+    // mistake. Checking this upfront also guarantees that the switch statements below only
+    // encounter labels they can handle.
+    if (std::find(labels.begin(), labels.end(), SorterType::ENERGY) != labels.end()) {
```
seweber, line 309: Also here I wonder whether it is necessary to handle ENERGY differently. If it is necessary, I wonder whether ENERGY could be excluded from SorterType. Currently, I have the impression that the code fluctuates between considering energy as a valid sorting criterion and considering it invalid.

src/cpp/src/basis/BasisPairCreator.cpp
```diff
@@ -272,8 +272,7 @@ std::shared_ptr<const BasisPair<Scalar>> BasisPairCreator<Scalar>::create() cons
     }
 
     // TODO: on the long run, construct the coefficient matrix directly
```
seweber, line 274: Is the TODO still required?

src/cpp/include/pairinteraction/interfaces/SorterBuilderInterface.hpp
```diff
@@ -0,0 +1,42 @@
...
+    std::set<size_t> boundaries;
+};
+
+class SorterBuilderInterface {
+public:
+    virtual ~SorterBuilderInterface() = default;
+    virtual Eigen::PermutationMatrix<Eigen::Dynamic, Eigen::Dynamic>
+    get_sorter(const std::vector<SorterType> &labels) const = 0;
+    virtual std::vector<IndicesOfBlock>
+    get_indices_of_blocks(const std::vector<SorterType> &labels) const = 0;
```
seweber, line 40: The interface has a bunch of purely virtual methods and is basically not implementing anything by its own. Moreover, only basis makes true use of the interface (system is just handing functions calls down to basis, if I see it correctly). Thus, I wonder whether we could remove this interface and implement everything in basis and delete get_sorter etc. from system?

src/cpp/src/system/System.cpp
```diff
@@ -74,36 +74,34 @@ System<Derived>::get_matrix() const {
 }
 
 template <typename Derived>
-const Transformation<typename System<Derived>::scalar_t> &
-System<Derived>::get_transformation() const {
+Eigen::PermutationMatrix<Eigen::Dynamic, Eigen::Dynamic>
+System<Derived>::get_sorter(const std::vector<SorterType> &labels) const {
+    // The first label of "labels" is supposed to be the primary sorting criterion.
```
seweber: label of labels?

src/cpp/src/system/System.cpp
seweber, line 1: Ah, I think now I remember where the special handling of ENERGY and sorting being also part of system comes from: energy is not yet known to the basis. Probably, my wishes are a topic for another PR after we have changed basis.

seweber (approved):

src/cpp/src/system/System.cpp
johannes-moegerle: changed to `The first sorter type in labels is supposed to be the primary sorting criterion.`

src/cpp/src/basis/BasisPairCreator.cpp
johannes-moegerle: the todo is about directly constructing the coefficient matrix of the basis, instead of applying a transformation to the basis, which we did not touch in this PR.
Whether this is a "problem" at all, and we want to change this at some point I am not sure

src/cpp/src/basis/Basis.cpp
johannes-moegerle: will be fixed by adding energy_of_states to the basis

src/cpp/src/basis/Basis.cpp
johannes-moegerle: will be fixed by adding energy_of_states to the basis

src/cpp/include/pairinteraction/interfaces/SorterBuilderInterface.hpp
johannes-moegerle: probably you are right, I will try to do this in a future PR

src/cpp/src/system/System.cpp
johannes-moegerle: yes, indeed, I after this PR I will directly try to make a PR adding energy_of_states to basis and hopefully all the ugly extra handling of energy is then gone

src/cpp/src/basis/BasisPairCreator.cpp
johannes-moegerle: will be adressed by a new constructor of the basis (that takes coefficients and good quantum numbers of the states)

PR #450, opened 2026-07-29
Gui improvements

johannes-moegerle:


src/pairinteraction/state/state_pair.py
```diff
@@ -93,7 +96,12 @@ def get_label(
 
         coeffs = np.abs(self.get_coefficients())
         ket_pair = self.get_ket(int(np.argmax(coeffs)))
-        ket_atom_tuple = tuple(state_atom.get_corresponding_ket() for state_atom in ket_pair.state_atoms)
+
+        # manually find the corresponding kets of the state atoms, to avoid warning messages
```
seweber, line 100: I wonder whether avoiding warning messages is really helpful here.

tests/test_gui.py
```diff
@@ -197,6 +197,24 @@ def _test_calculate_page(
...
+
+    for _ in range(3):  # further autosave ticks without user interaction must not touch the file
+        assert not base_window.save_settings()
+
+    ket_qn.items["n"].setValue(123)
+    assert base_window.save_settings()
+    assert base_window.settings_manager.value("OneAtomPage/ket_config/atom0_n", value_type=int) == 123
+
+    base_window.close()
+    plt.close("all")  # make sure to close all matplotlib figures
```
seweber: Why is `plt.close("all")` needed?

seweber (approved):

src/pairinteraction/state/state_pair.py
johannes-moegerle: this warning can fire, even though below this the heuristic is proven to have converged up to the desired included contributions.
See `is_converged` and the `logger.warning(` below, so the warnings messages from using `get_corresponding_ket` are only confusing and not real imo

tests/test_gui.py
johannes-moegerle: otherwise during pytest to many figures are "opened" and matplotlib will print a warning about it

seweber: I would have expected that base_window.close() should also take care of closing the figures, as they are part of the GUI, or am I overlooking something?

johannes-moegerle: fixed

PR #456, opened 2026-08-27
Small fixes

johannes-moegerle:


AGENTS.md
```diff
@@ -2,37 +2,60 @@
...
+uvx pre-commit run --files $(git diff --name-only HEAD) $(git ls-files --others --exclude-standard)
+```
+
+## Container limits
+
+If you are running inside a container, the host's CPU and memory figures overstate what you may actually use.
+
+Read the CPU and memory the container may actually use from the cgroup.
+Each value falls back to the host figure when the cgroup file is absent
+(cgroup v1, non-Linux) or reports `max`, and clamps to 1 when the limit is below one unit:
```
seweber: I am not sure whether I understand this sentence. What does "reports `max`" mean and "below one unit"?

seweber (approved):

PR #460, opened 2026-09-08
bugfix database expectation value range

johannes-moegerle:


src/pairinteraction/basis/basis_atom.py
```diff
@@ -72,6 +72,10 @@ def __init__(  # noqa: C901, PLR0912
     ) -> None:
         """Create a basis for a single atom.
 
+        All restrictions are inclusive and are widened by a small numerical tolerance, so that states are
```
seweber, line 75: I would not hardcode the tolerances in C++, but instead add them on the Python side. I am also not sure which tolerance values are appropriate or whether relative or absolute tolerances are better. My feeling is that a tolerance should cover the floating-point imprecision that could add up if one calculates a quantum number with MQDT - especially if the quantum number is supposed to be an integer. However, for things that are intrinsically floating-point values, such as energies or some quantum numbers in an MQDT setting, I would not use a tolerance. For these quantities, we anyway have an uncertainty from the uncertainty in physical parameters. The uncertainty coming from floating-point arithmetic is small in comparison to that.

seweber (changes requested):

src/pairinteraction/basis/basis_atom.py
johannes-moegerle: just as reference: `format_quantum_number_range` also has a tolerance, because this affects nu, which for high l should again be an integer (and probably is exactly an integer, but I am not entirely sure if one always can trust on that)

seweber (approved):

PR #464, opened 2026-09-15
fix download database

johannes-moegerle:


src/cpp/src/database/ParquetManager.cpp
seweber, line 1: Why was the last-modified mechanism replaced by etag? In principle, it is fine with me; I just wondered. Moreover, with etag, can we get rid of the "dummy authorization header" hack?

src/cpp/src/database/ParquetManager.test.cpp
```diff
@@ -25,15 +26,17 @@ class MockDownloader : public GitHubDownloader {
...
-            asset["name"] = "misc_v1.2.zip";
-            asset["url"] = "https://api.github.com/test/path/misc_v1.2.zip";
-            assets.push_back(asset);
-            nlohmann::json response;
-            response["assets"] = assets;
-            result.body = response.dump();
+            // This is a repo path request for a single release, return JSON with assets
+            result.body = make_release("1.2").dump();
+        } else if (remote_url == "/test/repo/releases") {
+            // This is a repo path request for a list of releases. The latest release requires a
```
seweber: This sounds a bit imprecise: The software requires a release with the exact major version specified in the software. I thus do not understand "requires a newer version of the software".

seweber (changes requested):

seweber (approved):

seweber (approved):

